PR #3, opened 2019-04-13
Update to conllx 0.11

danieldk:
Verified with topological field labeling on our standard splits.


sticker-utils/src/bin/sticker-prepare.rs
```diff
@@ -55,7 +55,7 @@ fn main() {
         .or_exit("Cannot relativize paths in configuration", 1);
 
     let input = Input::from(matches.free.get(1));
-    let treebank_reader = conllx::Reader::new(
+    let treebank_reader = Reader::new(
         input
             .buf_read()
             .or_exit("Cannot open corpus for reading", 1),
```
sebpuetz, line 61: Not necessarily connected to this PR or even this project: We are (almost) always using `or_exit()` for error-handling in `main()`. It's possible to change the signature of that method to `main() -> Result<(), Error>` and use the `?` in `main()` to handle errors. If something is throwing an error, the `Debug` representation of the `Error` is printed.

Sometimes having `or_exit` is the sole reason for the `stdinout` dependency (I know, `Output` and `Input` are quite handy, too). The difference between `or_exit` and returning the `Result` from main is that we're able to define a string message that gets displayed before the error message, but that message almost always ends up as some generic (often copy-pasted) message that doesn't add too much to debugging.

sebpuetz (approved):
Looks fine to me

sticker-utils/src/bin/sticker-prepare.rs
danieldk: I think I ranted about this in the Rust course. `Debug`-based errors printed when returning an error value from main are **really** ugly. (Or at least were when I last checked.) Fine for debugging, but not something that you wanted to throw at a user.

As for the generic error messages, we should use the `context` method provided by the `failure` crate more. Now we often just drop 'cannot open file' on the user. What file? 😉

PR #4, opened 2019-04-14
Make the noop and Tensorflow collectors generic

danieldk:
Both collectors assumed that the labels were of type String. This
change introduces the SentenceEncoder trait that encodes the tokens.
The collected type is an associated type of SentenceEncoder.

### DIFF
--- a/sticker/src/encoder.rs
+++ b/sticker/src/encoder.rs
@@ -1,0 +1,43 @@
+use conllx::graph::{Node, Sentence};
+use failure::{format_err, Error};
+
+use crate::{Layer, LayerValue};
+
+/// Trait for sentence encoders.
+///
+/// A sentence encoder extracts a representation of each token in a
+/// sentence, such as a part-of-speech tag or a topological field.
+pub trait SentenceEncoder {
+    type Encoding;
+
+    /// Encode the given sentence.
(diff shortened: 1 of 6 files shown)


sticker/src/encoder.rs
```diff
@@ -0,0 +1,48 @@
...
+        LayerEncoder { layer }
+    }
+}
+
+impl SentenceEncoder for LayerEncoder {
+    type Encoding = String;
+
+    fn encode(&self, sentence: &Sentence) -> Result<Vec<Self::Encoding>, Error> {
+        let mut encoding = Vec::with_capacity(sentence.len() - 1);
+        for node in sentence {
```
sebpuetz: Probably just personal preference but I like:
```Rust
for token in sentence.iter().filter_map(Node::token) {
...
}
```

sebpuetz (approved):
I like the changes, also from the clause-relations perspective since it's easy to implement an encoder to collect both fields and relations, too.

sticker/src/encoder.rs
danieldk: Agreed!

danieldk: Fixed, will merge when CI is happy.

PR #5, opened 2019-04-14
Make the Tag trait and the Tensorflow Tagger implementation generic

danieldk:
The Tagger type had `Numberer<String>`, replace by `Numberer<T>`. Also
provide a `SentenceDecoder` trait with an implementation for
`LayerEncoder`. This is now used by `SentProcessor` to update the
sentences.

### DIFF
--- a/sticker/src/encoder.rs
+++ b/sticker/src/encoder.rs
@@ -1,7 +1,21 @@
+use std::borrow::Borrow;
+
 use conllx::graph::{Node, Sentence};
 use failure::{format_err, Error};
 
 use crate::{Layer, LayerValue};
+
+/// Trait for sentence encoders.
+///
+/// A sentence encodes a representation of each token in a sentence,
+/// such as a part-of-speech tag or a topological field.
+pub trait SentenceDecoder {
+    type Encoding;
+
+    fn decode<E>(&self, labels: &[E], sentence: &mut Sentence) -> Result<(), Error>
+    where
+        E: Borrow<Self::Encoding>;
+}
 
 /// Trait for sentence encoders.
 ///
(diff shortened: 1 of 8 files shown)


sticker/src/encoder.rs
```diff
@@ -1,8 +1,22 @@
...
+/// Trait for sentence encoders.
+///
+/// A sentence encodes a representation of each token in a sentence,
+/// such as a part-of-speech tag or a topological field.
+pub trait SentenceDecoder {
+    type Encoding;
+
+    fn decode<E>(&self, labels: &[E], sentence: &mut Sentence) -> Result<(), Error>
+    where
+        E: Borrow<Self::Encoding>;
```
sebpuetz, line 17: Labels being a borrow of `Self::Encoding` seems somewhat counterintuitive and to my understanding requires an owner to borrow `E`. Currently there's the `Numberer` that provides the borrows which allows putting off allocation until the value is set on the token. There's not really much gained by putting off the allocation in my opinion.

I'm not sure if this would interfere with say binary classification where there's no numberer that provides borrows of the values.

sebpuetz (approved):
Looks good to me, just wondering whether `Decode` returning `Borrow`s of `Encoding` instead of `Encoding`s is the right choice.

sticker/src/encoder.rs
danieldk: But in other decoders, not the whole struct is always used. E.g. in one of the dependency parser encoders, only one struct field is used, the other two fields are only used to determine the head.

> I'm not sure if this would interfere with say binary classification where there's no numberer that provides borrows of the values.

Note that decode also takes owned labels, they just need to borrowable (`Borrow` has an impl for any `T`).

danieldk: Wanted to add another thing to the first point: in the dependency decoders we'll actually want to return the n-best labels, so the encoding could actually be a `Vec` of the actual encoding. Here it pays of even more to borrow.

PR #7, opened 2019-04-15
tensorflow: Add top-k tagging

danieldk:


sticker/src/tensorflow/tagger.rs
```diff
@@ -283,10 +304,46 @@ where
...
+            let seq_len = min(max_seq_len, sentence.borrow().len());
+            let offset = idx * max_seq_len * k;
+            let seq = &tag_tensor[offset..offset + seq_len * k];
+
+            labels.push(
+                seq.iter()
+                    .chunks(k)
+                    .into_iter()
+                    .map(|c| c.map(|&label| self.labels.value(label as usize).unwrap()))
+                    .map(FromIterator::from_iter)
```
sebpuetz: Could be more verbose (and less flexible) as: `.map(Vec::from_iter)`, don't think it's necessary to change but might be easier for code readers.

sebpuetz (approved):
Not really anything to complain about. Although, `Result<Vec<Vec<Vec<&T>>>, Error>` does look interesting as a return type ;)

sticker/src/tensorflow/tagger.rs
danieldk: Sounds good.

PR #8, opened 2019-04-16
Add the relative position dependency encoder

danieldk:
This encoder encodes the head as a relative position to a (dependent)
token.

This the initial version of the encoder. The decoding makes the attachment that it can make, but does not handle unattached tokens, or fix cycles. But I want to get all the machinery in place before making those refinements.


sticker/src/depparse/errors.rs
```diff
@@ -0,0 +1,29 @@
...
+    MissingRelation,
+}
+
+/// Decoder errors.
+#[derive(Clone, Copy, Debug, Eq, Fail, PartialEq)]
+pub enum DecodeError {
+    /// The head position is out of bounds.
+    #[fail(display = "position out of bounds")]
+    PositionOutOfBounds,
+
```
sebpuetz: Wondering if it ever matters for debugging whether out of bounds happened because of exceeding lower or upper bounds. If so, it might be nice to distinguish that either through another variant or a string message.

sebpuetz (approved):
14,605 added lines look somewhat intimidating.

Didn't catch anything worth mentioning here.

sticker/src/depparse/errors.rs
danieldk: I guess it might be useful. But currently, this particular enum variant is never actually exposed, but used by the decoder to move to the next most likely encoding if the first encoding cannot be applied.

Maybe the variants of this enum that don't bubble up should be put in a separate private enum.

danieldk: Actually, I think this whole enum can be crate-private.

danieldk: It's private now.

sebpuetz: Now I'm wondering why the `decode` methods return `Result`s at all. As far as I can tell, the following loop doesn't even return `Error`.

```Rust
            for encoding in encodings.as_ref() {
                if let Ok(triple) =
                    RelativePositionEncoder::decode_idx(idx, sentence.len(), encoding.borrow())
                {
                    sentence.dep_graph_mut().add_deprel(triple);
                    break;
                }
            }
```

Also, I think there's the edge case where the top k encodings are errors which leads to no deprel for the token?

danieldk: > Now I'm wondering why the `decode` methods return `Result`s at all. As far as I can tell, the following loop doesn't even return `Error`.

To give some leeway to decoders that may have failure conditions. If we can't find them, we could make it return void.

> Also, I think there's the edge case where the top k encodings are errors which leads to no deprel for the token?

Yes, there are at least three interacting edge cases (with some interaction):

1. Unattached tokens
2. Cycles
3. None of the tokens are attached to ROOT. I think this is only true if (1) or (2) is true.

But this PR and #9 just put the baseline encoder in place. PR #10 will add a nice utility for diagnostics. PR #10 will add support for parsing to `sticker-prepare`, `sticker-train`, and `sticker-tag` to support parsing as well.

PRs after that will focus on post-processing rules for fixups. I am currently implementing the fixups suggested by the paper, though with some modifications:

- The paper only uses top-1 (top-k is only used when no root could be found) and I use top-k as a back-off. I still have to evaluate whether my strategy gives better results.
- The post-processing in the paper reduces the number of tokens attached to ROOT to 1. I have some recollection that TüBa-D/Z has multiple tokens attached to ROOT in some cases, but I may misremember.

sebpuetz: > But this PR and #9 just put the baseline encoder in place. PR #10 will add a nice utility for diagnostics. PR #10 will add support for parsing to sticker-prepare, sticker-train, and sticker-tag to support parsing as well.

Am on #9 now.

> The post-processing in the paper reduces the number of tokens attached to ROOT to 1. I have some recollection that TüBa-D/Z has multiple tokens attached to ROOT in some cases, but I may misremember.

I remembered a [comment form Patricia](https://github.com/sfb833-a3/finalfrontier/pull/46#pullrequestreview-159293734) on one of my DepIter Pull requests

PR #10, opened 2019-04-16
Add sticker-dep2label: convert dependency relations into labels

danieldk:
This utility is for inspecting the generated labels. It adds them in a 'stringified' representation to the CoNLL-X features field. This utility is not actually used during training/prediction.

### DIFF
--- a/sticker-utils/src/bin/sticker-dep2label.rs
+++ b/sticker-utils/src/bin/sticker-dep2label.rs
@@ -1,0 +1,95 @@
+use std::env::args;
+use std::io::BufWriter;
+use std::process;
+
+use conllx::graph::Node;
+use conllx::io::{ReadSentence, WriteSentence};
+use conllx::token::Features;
+use getopts::Options;
+use stdinout::{Input, OrExit, Output};
+
+use sticker::depparse::{RelativePOSEncoder, RelativePositionEncoder};
+use sticker::SentenceEncoder;
+
+fn print_usage(program: &str, opts: Options) {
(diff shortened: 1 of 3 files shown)


sticker-utils/src/bin/sticker-dep2label.rs
```diff
@@ -0,0 +1,95 @@
...
+            .encode(&sentence)
+            .or_exit("Cannot dependency-encode sentence", 1);
+
+        for (token, encoding) in sentence.iter_mut().filter_map(Node::token_mut).zip(encoded) {
+            let mut features = token
+                .features()
+                .map(Features::as_map)
+                .cloned()
+                .unwrap_or_default();
+            features.insert("dl".to_owned(), Some(encoding.to_string()));
```
sebpuetz: I'm assuming `dl` is short for `dependency_label`? Looks a bit cryptic like this, maybe `dep_enc` or something?

sebpuetz (approved):
Not very transparent feature name, rest should be fine.

More general note: I'm not a fan of `getopts`, it's never obvious from reading the code what the commandline arguments are.

sticker-utils/src/bin/sticker-dep2label.rs
danieldk: I propose `deplabel` ;).

danieldk:
> More general note: I'm not a fan of `getopts`, it's never obvious from reading the code what the commandline arguments are.

I used `getopts` before `clap`, so some older stuff is still on `getopts`. `getopts` in C is even worse ;).

PR #11, opened 2019-04-17
Support parsing in all the utilities

danieldk:
For `sticker-prepare` and `sticker-train`, this is pretty straightforward: change from concrete types to `SentenceEncoder` and `SentenceEncoder::Encoding`. The changes to `sticker-tag` and `sticker-server` (and thus `sent_proc`) are a bit uglier. For the parser decoders, we need top-k results, whereas for layer decoding we only need the best results.

I considered using top-k results everywhere and just using the first results for the layer decoder. But this probably adds some unnecessary overhead, since it requires that a `Vec` needs to be constructed for every token.

I also tried to abstract the decoder interface completely to support both a slice of encodings or a slice of a slice of encodings. But this resulted in ugly abstractionitis.

Any opinions? I am for either this code duplication or returning top-k results everywhere, but not in favor of the abstractionitis approach.

### DIFF
--- a/sticker-utils/src/bin/sticker-prepare.rs
+++ b/sticker-utils/src/bin/sticker-prepare.rs
@@ -1,5 +1,7 @@
 use std::env::args;
 use std::fs::File;
+use std::hash::Hash;
+use std::io::Write;
 use std::path::Path;
 use std::process;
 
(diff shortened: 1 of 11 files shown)


danieldk:
I have to add that overhead is probably dwarfed by other things. Currently, by far most of the time is spend in Tensorflow first and far behind in the second place lookup/calculation of embeddings.

danieldk:
The first alternative is implemented in #12 . I think this simplifies thing a lot, both in the library and the utilities, at the cost of some extra overhead.

sebpuetz:
These PRs have a bit more volume, I'll try to get through them over the next few days.

danieldk:
Ok, thanks! I would start with #12, I think this is the more elegant solution. (Though it still needs a small change to work with CRFs as well.)

PR #12, opened 2019-04-17
Support parsing in all the utilities

danieldk:
To simplify the implementation, the SentenceTopKDecoder trait is
renamed to SentenceDecoder. The old SentenceDecoder is removed.

This is the alternative solution mentioned in #11. It does remove quite some code ;).

### DIFF
--- a/sticker-utils/testdata/sticker.conf
+++ b/sticker-utils/testdata/sticker.conf
@@ -1,5 +1,5 @@
 [labeler]
-  layer = { feature = "tf" }
+  labeler_type = { sequence = { feature = "tf" } }
   labels = "sticker.labels"
   read_ahead = 10
 
@@ -31,7 +31,6 @@
   inputs_op = "prediction/model/inputs"
   tags_op = "prediction/model/tags"
   seq_lens_op = "prediction/model/seq_lens"
-  predicted_op = "prediction/model/predicted"
   top_k_predicted_op = "prediction/model/top_k_predicted"
   accuracy_op = "prediction/model/accuracy"
   loss_op = "prediction/model/loss"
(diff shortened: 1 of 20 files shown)


sticker-utils/testdata/sticker.conf
```diff
@@ -31,7 +31,6 @@
   inputs_op = "prediction/model/inputs"
   tags_op = "prediction/model/tags"
   seq_lens_op = "prediction/model/seq_lens"
```
sebpuetz, line 33: These paths are incorrect, not related to the PR, but a future PR could/should fix that ;)

sebpuetz (approved):
I think it's fine to always return the topK results. If we're only interested in the single best sequence it's possible to just set `k=1`, the `Vec` wrapping the result really shouldn't matter, maybe some magic even makes it disappear in that case.

Also, this PR was pretty simple to get through, was expecting some heavier stuff based on line-count!

sticker-utils/testdata/sticker.conf
danieldk: Yeah, they are always wrong, and it's really annoying. I made these configurable for when somebody wants to define their own graph with custom op names. In hindsight, it may have been a bad choice. Since, it creates problems for 99% of the users for something 1% of the users would be interested in. Maybe I should switch back to hardcoding the op names in the tensorflow module. If someone creates a custom graph, it will still work if they use the same op names. They'd probably modify the existing Python scripts anyway.

sebpuetz: It's not really important since there's the `.conf` with correct paths under `example/`.

Although if the paths of the ops are changed, chances are that the Rust code also changed through adding, e.g., an additional op. I don't think anyone modifying the graph/model would be done by simply editing the `.conf` anyways, so it should be fine to just put them in the `tensorflow::model` module or wherever the ops reside.

PR #13, opened 2019-04-17
Actually create multiple RNN layers when requested

danieldk:
Also:

- Ditch the unidirectional RNN that was always applied.
- Instead, default to two bidirectional RNN layers.

These changes make the defaults more in line with most literature on
bidi RNN sequence labeling.


sebpuetz (approved):
Hope I'm not missing something again, but I think this is fine.

danieldk:
> Hope I'm not missing something again, but I think this is fine.

Well, the other diff didn't really have the relevant function in its context ;). PRs are quite ok for reviewing changes, but not so good for verifying interactions with other code.

PR #14, opened 2019-04-18
Hardcode Tensorflow op names

danieldk:
At some point, op names were added to the sticker configuration to
allow customization. However, in practice, they are an annoyance ---
if ops are renamed or added, all existing configurations break. So
switch back to hardcoding the op names.

### DIFF
--- a/sticker-utils/src/config_tests.rs
+++ b/sticker-utils/src/config_tests.rs
@@ -2,7 +2,7 @@ use std::fs::File;
 
 use lazy_static::lazy_static;
 use ordered_float::NotNan;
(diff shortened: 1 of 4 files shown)


sebpuetz (approved):
Looks fine, although I'm wondering whether we need to keep `OpNames` around? Should be possible to just plug the string literals into the `load_graph` method.

danieldk:
> Looks fine, although I'm wondering whether we need to keep `OpNames` around? Should be possible to just plug the string literals into the `load_graph` method.

My motivation was to bundle them. Since they are a lot of constants, this pollutes the namespace less. I guess it's even fewer lines of code to put constants in a module.

PR #15, opened 2019-04-23
Post-processing rules: find the root and attach orphans

danieldk:
The first post-processing step finds the root token (the first token
attached to the pseudo-root). The root is (following Strzyz, 2019):

- the first token with the ROOT relation if it exists; otherwise
- the token that attaches to ROOT with the highest probability; otherwise
- the first token of the sentence.

In the second and first cases, the head of the selected root token is
updated.

The second post-processing step attaches all orphans to the root.

Since probabilities are required for finding the highest probability
root, the tagger now returns encodings with probabilities. This also
required changes to decoders to handle the new probability wrapper.


danieldk:
Whoops, wrong branch name. Doesn't matter that much ;).

sebpuetz (approved):

PR #16, opened 2019-04-24
Remove cycles from the dependency graph

danieldk:
Applies the strategy suggested by Strzyz et al., 2019 to attach the
leftmost token in a cycle to the first root token.


sticker/src/depparse/post_processing.rs
```diff
@@ -29,6 +31,48 @@ where
     }
 }
 
+/// Break cycles in the graph.
+///
+/// Panics when a token does not have a head.
+pub fn break_cycles(sent: &mut Sentence, root_idx: usize) {
```
sebpuetz, line 39: I think almost all other methods return `Error`s to indicate failures, why panic here? From what I can tell, the calling method  `decode` returns an `Error`, too.

sticker/src/depparse/post_processing.rs
```diff
@@ -29,6 +31,48 @@ where
...
+    loop {
+        let components = {
+            let digraph: &DiGraph<_, _> = (&*sent).into();
```
sebpuetz, line 42: This looks somewhat ugly, to say the least. Perhaps a `.graph()` method on `Sentence` would make for some nicer looking code? I don't remember if we discussed that in the issues at the `conllx-rs` repo.
```Rust
let graph = sent.graph();
```

or even more concise (untested, don't know if lifetimes are an issue here):
```Rust
let components = tarjan_scc(sent.graph())
                .into_iter()
                .filter(|c| c.len() > 1)
                .collect::<Vec<_>>();
```

sticker/src/depparse/post_processing.rs
```diff
@@ -29,6 +31,48 @@ where
...
+        }
+
+        for cycle in components.into_iter().filter(|c| c.len() > 1) {
+            // Find the first token in the cycle, exclude the root
+            // token to avoid self-cycles.
+            let first_token = cycle
+                .into_iter()
+                .filter(|idx| idx.index() != root_idx)
+                .min()
+                .expect("Cannot get minimum from a non-empty iterator")
```
sebpuetz: "from an empty iterator"?

sticker/src/depparse/post_processing.rs
```diff
@@ -29,6 +31,48 @@ where
...
+/// Break cycles in the graph.
+///
+/// Panics when a token does not have a head.
+pub fn break_cycles(sent: &mut Sentence, root_idx: usize) {
+    loop {
+        let components = {
+            let digraph: &DiGraph<_, _> = (&*sent).into();
+            tarjan_scc(digraph)
+                .into_iter()
+                .filter(|c| c.len() > 1)
```
sebpuetz, line 45: Is the same `filter` statement necessary on line 52, too?

sticker/src/depparse/post_processing.rs
```diff
@@ -29,6 +31,48 @@ where
...
+                .collect::<Vec<_>>()
+        };
+
+        // We are done if there are no more cycles.
+        if components.is_empty() {
+            break;
+        }
+
+        for cycle in components.into_iter().filter(|c| c.len() > 1) {
```
sebpuetz: `filter` seems redundant to me, line 43 contains the same condition and is executed in the same loop iteration.

sebpuetz (approved):
Added some comments about style, a typo and a possibly redundant `filter` call. Functionality should be fine, anyways.

sticker/src/depparse/post_processing.rs
danieldk: This is indeed missing in `conllx-rs`. Another annoyance is that `DepGraphMut` does not have a method to remove edges (something like `remove_deprel(dependent: usize)`). I'll add issues.

What makes things less ergonomic in this case is that type-inference does understand the simple `into()` call, because sent is `&mut` and the `From` implementation is for `&`. Not providing an `&mut` implementation was intentional to avoid changing the graph from under our feet.

sticker/src/depparse/post_processing.rs
danieldk: I did mean non-empty, because that is what the iterator is guaranteed to be at this point. Changed the error message a bit.

sticker/src/depparse/post_processing.rs
danieldk: Good catch.

sticker/src/depparse/post_processing.rs
danieldk: `encode`/`decode` return errors for invalid user input. This is a private function and having a head on each token is an invariant. Passing a graph with headless tokens to `break_cycles` is a programming error and should thus lead to an assertion-type panic and not to a regular error that is communicated to the user.

I have extended the `rustdoc` a bit that `attach_orphans` should be called before this function if the graph has headless tokens.

sticker/src/depparse/post_processing.rs
danieldk: https://github.com/danieldk/conllx-rs/issues/11
https://github.com/danieldk/conllx-rs/issues/10

sticker/src/depparse/post_processing.rs
sebpuetz: The rustdoc of `.min()` states that the method returns `None` if the iterator is empty. The iterator is guaranteed to be non-empty, but I guess, `.expect()` or `.unwrap()` don't make much of a difference here. I just didn't correctly parse the string argument.

sticker/src/depparse/post_processing.rs
sebpuetz: Didn't catch that the `post_processing` module was crate-private, I only looked at the `pub fn break_cycles` signature.

danieldk: Ok, I'll merge this then.

PR #22, opened 2019-05-02
Move the input length computation to SentVectorizer

danieldk:
This puts the input length computation where it belongs.

### DIFF
--- a/sticker/src/input.rs
+++ b/sticker/src/input.rs
@@ -91,6 +91,17 @@ impl SentVectorizer {
         SentVectorizer { layer_embeddings }
     }
 
+    /// Get the length of the input representation.
+    pub fn input_len(&self) -> usize {
+        self.layer_embeddings.token_embeddings().dims()
+            + self
+                .layer_embeddings
+                .tag_embeddings()
+                .as_ref()
+                .map(|e| e.dims())
+                .unwrap_or_default()
+    }
+
     /// Get the layer embeddings.
     pub fn layer_embeddings(&self) -> &LayerEmbeddings {
         &self.layer_embeddings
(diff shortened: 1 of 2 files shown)


sebpuetz (approved):
Looks fine. 
Just a few thoughts from my side:

*  `input_len()` is a name that I'd expect to refer to the number of elements in an input rather than the dimensionality of a vector representing a single element. In the context (and its usage) it should be clear what's meant anyways.
* I would expect that method on the `LayerEmbeddings`, since that's where all the information is located anyways.
* This is less related to this PR, but more generally relevant, can we gain something by abstracting over the `LayerEmbedding`? I know it's a thin line between too much abstraction and hacky implementation, but I'm wondering whether there should be distinct embedding structs for `Word`, `Word + POS`, which could make inputs more flexible if, e.g., pretrained character embeddings are added. This comes at the cost of having to implement seperate structs (and the corresponding trait) for each type, but I think the number of possible types should be rather limited.

danieldk:
> `input_len()` is a name that I'd expect to refer to the number of elements in an input rather than the dimensionality of a vector representing a single element. In the context (and its usage) it should be clear what's meant anyways.

For me that is the sequence length? But I can see why it is confusing.

> This is less related to this PR, but more generally relevant, can we gain something by abstracting over the `LayerEmbedding`? I know it's a thin line between too much abstraction and hacky implementation, but I'm wondering whether there should be distinct embedding structs for `Word`, `Word + POS`, which could make inputs more flexible if, e.g., pretrained character embeddings are added. This comes at the cost of having to implement seperate structs (and the corresponding trait) for each type, but I think the number of possible types should be rather limited.

Could be. I considered it when adding tags. Let's see what happens when characters are added. I think once more layers are added, I'd rather opt for the approach that I took in dpar and use an `EnumMap`.

PR #23, opened 2019-05-03
Use TensorBuilder in TensorCollector

danieldk:
In order to do so, extend TensorBuilder to also support collection of
labels.

### DIFF
--- a/sticker/src/tensorflow/tensor.rs
+++ b/sticker/src/tensorflow/tensor.rs
@@ -2,44 +2,126 @@ use std::cmp::min;
 
 use tensorflow::Tensor;
 
-pub struct TensorBuilder {
+mod labels {
+    pub trait Labels {
+        fn from_shape(batch_size: u64, time_steps: u64) -> Self;
+    }
+}
+
+/// No labels.
+#[derive(Default)]
+pub struct NoLabels;
+
+impl labels::Labels for NoLabels {
+    fn from_shape(_batch_size: u64, _time_steps: u64) -> Self {
+        NoLabels
+    }
+}
+
+/// Labels stored in a `Tensor<i32>`.
+pub struct LabelTensor(pub Tensor<i32>);
+
+impl labels::Labels for LabelTensor {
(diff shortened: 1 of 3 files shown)


sticker/src/tensorflow/tensor.rs
```diff
@@ -42,4 +74,34 @@ impl TensorBuilder {
...
+        &self.labels
+    }
+}
+
+impl TensorBuilder<LabelTensor> {
+    pub fn add_with_labels(&mut self, input: &[f32], labels: &[i32]) {
+        self.add_input(input);
+
+        let max_seq_len = self.inputs.dims()[1] as usize;
+        let token_embed_size = self.inputs.dims()[2] as usize;
```
sebpuetz: That's not only the dimensionality of the token embeddings but might also include other embeddings, right?

sticker/src/tensorflow/tensor.rs
```diff
@@ -31,8 +61,10 @@ impl TensorBuilder {
         let token_seq =
             &mut self.inputs[token_offset..token_offset + (token_embed_size * timesteps)];
         token_seq.copy_from_slice(&input[..token_embed_size * timesteps]);
```
sebpuetz: `token_embed_size` is somewhat ambiguous.

sebpuetz (approved):
Looks good to me, although there's some ambiguity with `token_embed_size`. Could be read as the size of the token embedding or the size of the embeddings associated with a token.

danieldk:
> Looks good to me, although there's some ambiguity with `token_embed_size`. Could be read as the size of the token embedding or the size of the embeddings associated with a token.

"There are only two hard things in Computer Science: cache invalidation and naming things" -- Phil Karlton

I have changed the name to `token_repr_size`. 'Token representation' is hopefully a bit less ambiguous than 'token embedding'.

danieldk:
Thanks for the review! Will merge once Travis-CI is happy.

PR #24, opened 2019-05-07
Switch tensors to NdTensor

danieldk:
Rust 1.31 is removed from CI builds, since ndarray-tensorflow uses
TryFrom. Bump to Rust 1.34.0, which stabilizes TryFrom.


sticker-utils/src/bin/sticker-train.rs
```diff
@@ -186,7 +186,7 @@ where
             tagger.validate(seq_lens, tokens, labels)
         };
 
-        let n_tokens = seq_lens.iter().sum::<i32>();
+        let n_tokens = seq_lens.view().iter().sum::<i32>();
```
sebpuetz, line 189: Seems odd/unexpected, that the `NDTensor` doesn't offer the .iter() method. Maybe `.iter()` and `.iter_mut()` can be implemented on `NDTensor`, too?

sebpuetz (approved):

sticker-utils/src/bin/sticker-train.rs
danieldk: Great idea! I guess that we'd also need axis iterators.

danieldk: Added an issue: danieldk/ndarray-tensorflow#2

PR #26, opened 2019-05-09
Make GLU the default activation for convolution networks

danieldk:


sebpuetz (approved):
There's some redundancy coming from `argparse`'s default params and additionally defining them in `DefaultConfig`. Leads to some confusion when changing values in `DefaultConfig` but not seeing any effect because the numbers are overwritten in the main method anyways.

danieldk:
> There's some redundancy coming from `argparse`'s default params and additionally defining them in `DefaultConfig`. Leads to some confusion when changing values in `DefaultConfig` but not seeing any effect because the numbers are overwritten in the main method anyways.

I agree, it's a leftover from when the hyperparameters were hardcoded.

PR #32, opened 2019-05-14
Split Tagger into Tagger and Trainer data structures

danieldk:
This reduces the dependency burden for training uses, since no labels
and vectorizer are needed.

### DIFF
--- a/sticker-utils/src/bin/sticker-train.rs
+++ b/sticker-utils/src/bin/sticker-train.rs
@@ -12,7 +12,7 @@ use indicatif::{ProgressBar, ProgressStyle};
 use stdinout::OrExit;
 use sticker::depparse::{RelativePOSEncoder, RelativePositionEncoder};
 use sticker::tensorflow::{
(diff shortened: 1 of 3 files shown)


sebpuetz (approved):
Looks fine, but I think `TaggerGraph` is not a very clear name for something used both in `Trainer` and `Tagger`. Also, the module `tagger` both holding the `Trainer` and `Tagger` seems odd.

danieldk:
> Looks fine, but I think `TaggerGraph` is not a very clear name for something used both in `Trainer` and `Tagger`. Also, the module `tagger` both holding the `Trainer` and `Tagger` seems odd.

Good points, I'll reshuffle them in the future.

danieldk:
> Good points, I'll reshuffle them in the future.

Maybe `Trainer` should be renamed to `TaggerTrainer`? That would explain the relations between the pieces better.

sebpuetz:
Sounds okay

PR #33, opened 2019-05-14
Generalize FileProgress to ReadProgress

danieldk:
Change the wrapper to work with any Read + Seek. Also implement Seek.


sebpuetz (approved):
Don't see anything to complain about, although I'm wondering where `set_style` is going to be used.

danieldk:
> Don't see anything to complain about, although I'm wondering where `set_style` is going to be used.

The next set of changes. Sticker training now uses too much memory and I would like to support pretraining scenarios, so the next few changes will switch to on-the-fly tensor generation. However, in that case we don't know the number of batches ahead of time (especially in pretraining, where the CoNLL-X data may be large). So, all the progress information will come from `ReadProgress`.

Issue #28, opened 2019-05-11
Labels from dev set are added to Numberer


sebpuetz:
If there are labels in the dev set that are not in the train set, the labels get added to the numberer, leading to `NaN`s on the validation losses. Only the display of the validation batch results is affected from this, so it doesn't really have any consequences.

There's the workaround to run `sticker-prepare` on the full corpus rather than just the train set, but that adds unseen and unused classes to the softmax calculation.

Perhaps set the starting value of the numberer higher and reserve a number for unseen labels?

danieldk:
A special value would indeed solve this. dpar has the distinction between mutable and frozen lookups, a frozen lookup can/could assign this value for any known value. I think the special value could even be the label for padding (0), since we carry around the actual sequence lengths, so we do not have to rely on this special label.

The `NaN` loss hasn't annoyed me enough yet to actually do something about this ;). Maybe after the other major changes are done.

sebpuetz:
Once you're aware of what's going on, it's not a problem anymore. On the other hand, spending time to figure out why the loss shows up as NaN is somewhat annoying.

danieldk:
That's true. As a temporary workaround, we could simply elide the loss for validation.

PR #34, opened 2019-05-15
Introduce CategoricalEncoder

danieldk:
This is a wrapper around a Sentence{Encoder,Decoder} and Numberer,
that directly encoder from/to usize categorical labels.

---

The motivation for this change was that we carry around encoders and labelers together, since they are typically used together: encode and map to numbers, map numbers to encodings and decode. So why not put them together. `CategoricalEncoder` wraps any encoder/decoder to directly go to/from numbers.

`CategoricalEncoder` is now also used directly in `Tagger`, avoiding the need to allocate the full intermediate `Vec<Vec<Vec<_>>>` of encodings with their probabilities.

### DIFF
--- a/sticker/src/encoder.rs
+++ b/sticker/src/encoder.rs
@@ -1,9 +1,10 @@
 use std::borrow::{Borrow, Cow};
+use std::hash::Hash;
 
 use conllx::graph::{Node, Sentence};
 use failure::{format_err, Error};
 
-use crate::{Layer, LayerValue};
+use crate::{Layer, LayerValue, Numberer};
 
 /// An encoding with its probability.
 pub struct EncodingProb<'a, E>
@@ -76,7 +77,7 @@ pub trait SentenceEncoder {
     type Encoding;
 
     /// Encode the given sentence.
-    fn encode(&self, sentence: &Sentence) -> Result<Vec<Self::Encoding>, Error>;
+    fn encode(&mut self, sentence: &Sentence) -> Result<Vec<Self::Encoding>, Error>;
 }
 
 /// Encode sentences using a CoNLL-X layer.
@@ -123,7 +124,7 @@ impl SentenceDecoder for LayerEncoder {
 impl SentenceEncoder for LayerEncoder {
     type Encoding = String;
 
-    fn encode(&self, sentence: &Sentence) -> Result<Vec<Self::Encoding>, Error> {
+    fn encode(&mut self, sentence: &Sentence) -> Result<Vec<Self::Encoding>, Error> {
         let mut encoding = Vec::with_capacity(sentence.len() - 1);
         for token in sentence.iter().filter_map(Node::token) {
             let label = token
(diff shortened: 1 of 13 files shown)


sticker/src/encoder.rs
```diff
@@ -135,3 +136,124 @@ impl SentenceEncoder for LayerEncoder {
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::fs::File;
+    use std::io::BufReader;
+    use std::path::Path;
+
+    use conllx::graph::{Node, Sentence};
```
sebpuetz: I think those are unused

sticker/src/tensorflow/tagger.rs
```diff
@@ -238,32 +238,38 @@ where
...
-        for (idx, sentence) in sentences.iter().enumerate() {
-            let seq_len = min(max_seq_len, sentence.borrow().len() - 1);
+        for (idx, sentence) in sentences.iter_mut().enumerate() {
+            // BorrowMut derives from borrow, but the borrowed type cannot
+            // be inferred with a borrow() call here?
+            let seq_len = min(
+                max_seq_len,
+                // Help the type checker, can't infer borrowed type otherwise.
+                <BorrowMut<Sentence> as Borrow<Sentence>>::borrow(sentence).len() - 1,
+            );
```
sebpuetz:
```Rust
let sentence = sentence.borrow_mut();
let seq_len = min(
    max_seq_len,
    sentence.len() - 1,
);
```
would make this more readable, `.borrow_mut()` on line 269 could then be removed, too.

sebpuetz: Or map the iterator:
```Rust
for (idx, sentence) in sentences.iter_mut().map(BorrowMut::borrow_mut).enumerate() {
    //[...]
}
```

sticker/src/encoder.rs
```diff
@@ -135,3 +136,124 @@ impl SentenceEncoder for LayerEncoder {
...
+impl<E> SentenceEncoder for CategoricalEncoder<E, E::Encoding>
+where
+    E: SentenceEncoder,
+    E::Encoding: Clone + Eq + Hash,
+{
+    type Encoding = usize;
+
+    fn encode(&mut self, sentence: &Sentence) -> Result<Vec<Self::Encoding>, Error> {
+        let encoding = self.inner.encode(sentence)?;
+        let categorical_encoding = encoding.into_iter().map(|e| self.numberer.add(e)).collect();
```
sebpuetz, line 170: Fixing #28 would require adding a flag or another method here, e.g.
```Rust
fn encode(&mut self, sentence: &Sentence, mode: Mode) -> Result<Vec<Self::Encoding>, Error> {
    let encoding = self.inner.encode(sentence)?;
    let categorical_encoding = match mode {
        Mode::Train => encoding.into_iter().map(|e| self.numberer.add(e)).collect()
        // would require starting the numberer at 2
        Mode::Validate => encoding.into_iter().map(|e| self.numberer.number(e).unwrap_or(1)).collect()
}
```

sebpuetz (approved):

sticker/src/encoder.rs
danieldk: I don't like adding arguments. I think this should rather be addressed by the `Numberer`, which could be thawed/frozen. (As dpar does it.)

danieldk:
Thanks for the suggestions! I've incorporated them, will merge when CI is happy.

PR #35, opened 2019-05-15
Splits write-graph into model specific scripts.

twuebi:
This splits the write-graph script into model specific scripts while factoring out common functionality.

With the addition of more architectures, the arguments become increasingly unclear. This PR replaces the existing `write-graph` scripts with `write-rnn` and `write-cnn`. Commonly used functionality has moved to `write_helper.py`.


sebpuetz (commented):
I don't see anything to complain about, but @danieldk probably wants to take a look, too.

danieldk:
> I don't see anything to complain about, but @danieldk probably wants to take a look, too.

Yep, you can assign me as a reviewer for sticker PRs. I will have a look at this (and try it out).

sticker-utils/tensorflow/write-cnn
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python
```
danieldk: I think this is also a good excuse to fully move to Python 3. Could you change the shebang to

```
#!/usr/bin/env python3
```

? And the `coding` line can be removed, because Python 3's default encoding is UTF-8.

sticker-utils/tensorflow/write-cnn
```diff
@@ -0,0 +1,50 @@
...
+#  -*- coding: utf-8 -*-
+
+from conv_model import ConvModel
+from write_helper import get_common_parser, parse_common_config, create_graph
+
+if __name__ == '__main__':
+    parser = get_common_parser()
+    # conv
```
danieldk: Spurious comment.

sticker-utils/tensorflow/write_helper.py
```diff
@@ -0,0 +1,66 @@
...
+    with open(args.shape_file) as shapesfile:
+        shapes = toml.loads(shapesfile.read())
+    return shapes
+
+
+def create_graph(config, model, args):
+    shapes = read_shapes(args)
+    graph_filename = args.output_graph_file
+
+    gpuopts = tf.GPUOptions(per_process_gpu_memory_fraction=0.3)
```
danieldk, line 18: Not for this PR, but in a future PR, we should change this to always run on the CPU. There is not much use running this on a GPU.

danieldk (changes requested):
First of all, thanks for doing this, the mixture of options was getting a bit convoluted.

Some small comments. Would it also be possible to rename the utilities to respectively `sticker-write-rnn-graph` and `sticker-write-conv-graph`?

The `graph` suffix because it states more explicitly what is being written. The `sticker-` prefix, because we may want to include these utilities in distribution packages as well. The `sticker-` prefix then makes them more discoverable using tab completion.

sticker-utils/tensorflow/write-cnn
twuebi: The model files also feature `#!/usr/bin/env python`. Is there any use to it?

danieldk: Nope, they are accidental side-effects of copy paste. They can definitely be removed.

twuebi: In this PR?

danieldk: Better to do it in a separate PR.

danieldk:
Could you squash the changes into one commit once the PR is ready to go? You've probably done it before, if not:

~~~
git rebase -i master
# set the latest commit to squash, save and exit
git push --force-with-lease
~~~

twuebi:
Done

PR #37, opened 2019-05-15
Add the DataSet trait and provide an implementation for CoNLL-X.

danieldk:
The DataSet trait provides a batches() method that returns an iterator
over (Tensorflow) batches of the dataset. Implementations of DataSet
should generate the batches on-the-fly to avoid excessive memory use.

The ConllxDataSet data structure implements DataSet for CoNLL-X.

---

I will open a separate PR for the integration of `DataSet`.

### DIFF
--- a/sticker/src/tensorflow/dataset.rs
+++ b/sticker/src/tensorflow/dataset.rs
@@ -1,0 +1,129 @@
+use std::hash::Hash;
+use std::io::{BufReader, Read, Seek, SeekFrom};
+
+use conllx::io::{ReadSentence, Reader, Sentences};
+use failure::Fallible;
+
+use super::tensor::{LabelTensor, TensorBuilder};
+use crate::{CategoricalEncoder, SentVectorizer, SentenceEncoder};
+
+/// A set of training/validation data.
+///
+/// A data set provides an iterator over the batches in that
+/// dataset.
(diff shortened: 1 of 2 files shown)


sebpuetz (approved):
Looks fine, although I don't quite understand why `DataSet` is implemented on `&'ds mut ConllxDataSet<R>` without more context. I guess it will get passed into `run_epoch` by reference rather than by value?

danieldk:
> Looks fine, although I don't quite understand why `DataSet` is implemented on `&'ds mut ConllxDataSet<R>` without more context.

It's  an interesting problem. I need the lifetime for the iterator (since we cannot consume the reader, otherwise we could only call `batches` once). However, the compiler objects that the lifetime `'ds` is not used if `DataSet` is implemented for `ConllxDataSet`, rather than `&ConllxDataSet`. I guess it makes sense, because the associated type `Iter` is only properly defined with the instantiation of that lifetime. I guess that the lifetime cannot be properly defined for `ConllxDataSet`, since such a value might move.

A close analogy is `IntoIterator` of slices/string references in the standard library (since the corresponding iterator also borrows), which also have `IntoIterator` implementations on the reference types `&[T]`/`&str` rather than the unsized types `[T]`/`str`.

PR #38, opened 2019-05-16
sticker-train: switch to the DataSet API

danieldk:
This make memory use during training O(1) rather than O(n), where n is
the corpus size.


sticker-utils/src/bin/sticker-train.rs
```diff
@@ -118,113 +176,59 @@ fn train_model(
...
 }
 
-fn run_epoch(
+fn run_epoch<E>(
+    config: &Config,
+    encoder: &mut CategoricalEncoder<E, E::Encoding>,
+    vectorizer: &SentVectorizer,
     trainer: &TaggerTrainer,
-    tensors: &CollectedTensors,
+    file: &mut File,
```
sebpuetz, line 184: Reference to the file is kept instead of re-opening the file after each epoch, so external changes to the file don't mess up the training?

sticker-utils/src/bin/sticker-train.rs
```diff
@@ -118,113 +176,59 @@ fn train_model(
...
-        let labels = &tensors.labels[i];
+    for batch in dataset
+        .batches(encoder, vectorizer, config.model.batch_size)
+        .or_exit("Cannot read batches", 1)
+    {
+        let (inputs, seq_lens, labels) = batch.or_exit("Cannot read batch", 1).into_parts();
 
         let batch_perf = if is_training {
-            trainer.train(seq_lens, tokens, labels, lr)
+            trainer.train(&seq_lens, &inputs, &labels.0, lr)
```
sebpuetz, line 214: Maybe using a named field rather than `.0` could make this clearer?
Or define a function on `LabelTensor` like `tensor()` returning a ref to the inner tensor as `labels.tensor()`?

sebpuetz (approved):
Looks fine.

Although not directly related, but there is an alternate solution to implementing `DataSet` for a mutable reference:

Change `DataSet::batches` to be non-consuming and add the `'ds` lifetime as a parameter

```Rust
pub trait DataSet<'a, 'ds, E>
where
    E: SentenceEncoder,
    E::Encoding: Clone + Eq + Hash,
{
    type Iter: Iterator<Item = Fallible<TensorBuilder<LabelTensor>>>;

    fn batches(
        &'ds mut self,
        encoder: &'a mut CategoricalEncoder<E, E::Encoding>,
        vectorizer: &'a SentVectorizer,
        batch_size: usize,
    ) -> Fallible<Self::Iter>;
```

the impl then reads as

```Rust
impl<'a, 'ds, E, R> DataSet<'a, 'ds, E> for ConllxDataSet<R>
where
    E: 'a + SentenceEncoder,
    E::Encoding: 'a + Clone + Eq + Hash,
    R: Read + Seek + 'ds,
{
    type Iter = ConllxIter<'a, E, Reader<BufReader<&'ds mut R>>>;
```

danieldk:
> Although not directly related, but there is an alternate solution to implementing `DataSet` for a mutable reference:

Right, that's another solution, but then you can't make implementations where calling `batches` consumes the data set. This may be useful for cases where you cannot seek on the underlying reader. Don't know if we ever need it though ;).

sticker-utils/src/bin/sticker-train.rs
danieldk: Depends on the kind of changes (does it change the file in-place or not). But at least on unixes, you can remove or move the file and the file descriptor will continue to be valid as long as it's open.

sticker-utils/src/bin/sticker-train.rs
danieldk: I'd be in favor of a named field.Maybe we don't need the `LabelTensor` wrapper at all, I guess we could just directly implement the `Labels` trait for `NdTensor<i32, Ix2>`. I'll do a separate PR for this.

PR #39, opened 2019-05-16
Remove the LabelTensor wrapper struct

danieldk:
@sebpuetz something like this?

Not too sure if the type alias adds something. Maybe it makes the impls easier to read?

### DIFF
--- a/sticker-utils/src/bin/sticker-train.rs
+++ b/sticker-utils/src/bin/sticker-train.rs
@@ -211,9 +211,9 @@ where
         let (inputs, seq_lens, labels) = batch.or_exit("Cannot read batch", 1).into_parts();
 
         let batch_perf = if is_training {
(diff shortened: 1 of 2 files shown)


sebpuetz (approved):
I'd be happy with that. 

Quickly reading through code and seing `.0` on `labels` just isn't very telling about what the `.0` is doing there. To figure that out, the reader would have to go to the `DataSet` impl and from there to `LabelTensor` to figure out that this just gets the wrapped component from a tuple struct.

danieldk:
> I'd be happy with that.
> 
> Quickly reading through code and seing `.0` on `labels` just isn't very telling about what the `.0` is doing there. To figure that out, the reader would have to go to the `DataSet` impl and from there to `LabelTensor` to figure out that this just gets the wrapped component from a tuple struct.

Fully agree!

PR #45, opened 2019-05-17
Convert sticker-graph to a proper Python package

danieldk:
This change converts the Python code into something that is `pip install`-able. I am not a Python expert, let alone a Python packaging export. Maybe @twuebi could also have a look, since you seem to be the closest thing to a Pythonista in our group ;).


sticker-graph/setup.py
```diff
@@ -0,0 +1,23 @@
+from setuptools import setup
+
+setup(name='sticker-graph',
+      version='0.2.0',
+      description='Graph construction utilities for stickerd',
```
sebpuetz: Typo?

danieldk: Oops. Thanks!

sticker-graph/setup.py
```diff
@@ -0,0 +1,23 @@
...
+      author='Daniël de Kok',
+      author_email='me@danieldk.eu',
+      license='Apache2',
+      tests_require=[
+          'numpy',
+          'tensorflow == 1.13.1',
+          'toml',
+      ],
+      install_requires=[
+          'tensorflow == 1.13.1',
```
twuebi, line 16: `python setup.py install` starts downloading `tensorflow-cpu` via pip even if one of`tensorflow-gpu` or `tensorflow-rocm` is present. According to this [so-post](https://stackoverflow.com/a/49338206), something along these lines could avoid this:
```python
from pkg_resources import DistributionNotFound, get_distribution

def get_dist(pkgname):
    try:
        return get_distribution(pkgname)
    except DistributionNotFound:
        return None


install_requires = []
if get_dist('tensorflow') is None and get_dist('tensorflow_gpu') is None and get_dist('tensorflow_rocm') is None:
    install_requires.append('tensorflow')


setup(name='sticker-graph',
      ...
      tests_require=[
                        'numpy',
                        'toml',
                    ] + install_deps,
      install_requires=[
                           'toml',
                       ] + install_deps,
      ...
      )


```

sticker-graph/sticker_graph/__init__.py
```diff
@@ -0,0 +1,3 @@
+from sticker_graph.conv_model import ConvModel
```
twuebi: These are unused, in the `write-scripts` the imports are absolute. Is this intended for possible library use? In that case, we should provide more pydoc on the `shapes` and `config` parameters of the models.

sticker-graph/sticker_graph/conv_model.py
```diff
@@ -1,9 +1,8 @@
+from enum import Enum
```
twuebi, line 1: Minor: blank line between standard library and other imports

sticker-graph/sticker_graph/rnn_model.py
```diff
@@ -1,7 +1,7 @@
 import tensorflow as tf
 from tensorflow.contrib.layers import batch_norm
 from enum import Enum
```
twuebi: Unused import

sticker-graph/sticker_graph/conv_model.py
```diff
@@ -1,9 +1,8 @@
...
 import tensorflow as tf
 from tensorflow.contrib.layers import batch_norm
```
twuebi, line 4: (not part of this PR:) unused import

twuebi (changes requested):
The `setup.py` downloads `tensorflow-cpu` if not present. Besides that, some minor things.

sticker-graph/sticker_graph/__init__.py
danieldk: > These are unused, in the write-scripts the imports are absolute.

No? E.g. in `sticker-write-rnn-graph`:

~~~python
from sticker_graph import RNNModel
~~~

Or am I missing something (very much possible ;))?

sticker-graph/sticker_graph/conv_model.py
danieldk: I am fine with that, but is there a standard? `autopep8` does not enforce a blank line.

sticker-graph/sticker_graph/conv_model.py
danieldk: Ok, will do that in a separate PR.

sticker-graph/sticker_graph/rnn_model.py
danieldk: Thanks!

sticker-graph/sticker_graph/conv_model.py
danieldk: Fixed, thanks!

sticker-graph/setup.py
danieldk: But I want the CPU version ;), running a GPU version of Tensorflow to write the graph is a waste. I don't really like the SO solution, because then we have to maintain a list of different Tensorflow versions out there. E.g., there is also `intel-tensorflow` (which is probably compiled with Intel MKL-DNN).

At any rate, we should encourage that people use per-project virtual environments, which is especially easy since Python has included the `venv` module. And since the training is done in Rust-land, the CPU version is of Tensorflow is the correct dependency for a sticker venv.

sticker-graph/sticker_graph/__init__.py
twuebi: Oh, my bad. I checked `rnn_model.py` and `conv_model.py` for imports and mistook the absolute imports of `Model` there. 

This seems fine then. Although, it still invites to (undocumented) library use.

danieldk: Indeed, it does. I don't know if we actually want to provide this as a public API. My motivation for separating the module was so that we can make distro packages, where the scripts have to live in `$PREFIX/bin` and the remaining Python files in some module directory.

Maybe we could add some remark that the API is not documented, because it is not intended to be a public API?

sticker-graph/sticker_graph/conv_model.py
twuebi: Afaik, there are various standards. Tensorflow seems to do:

```
future

standard lib

external

local
```

I like organizing them in these blocks as it separates the several levels.

danieldk: I like it! That's my Rust ordering (except for future).

sticker-graph/setup.py
twuebi: I'd still prefer not downloading an unnecessary dependency, but I guess it's not that important.

sticker-graph/sticker_graph/__init__.py
twuebi: A start could be to remove the imports from `__init__.py` and replace the ones in the script by absolute imports. Would this clash with one of your envisioned use cases?

sticker-graph/setup.py
danieldk: Allow me to brag a little ;):

~~~bash
~ % nix path-info nixpkgs.python3Packages.tensorflow
/nix/store/fn697pz7j69zwjn703nmxcym3bdfhps8-python3.7-tensorflow-1.13.1
~ % nix path-info nixpkgs.python3Packages.tensorflowWithCuda
/nix/store/za3w1py46mfwf5r0g1fr7arrj6f07njb-python3.7-tensorflow-1.13.1
~~~

Shared between all my Tensorflow projects.

~~~bash
$ nix-shell -p python3Packages.tensorflow
~~~

and I am in a shell with CPU Tensorflow,

~~~bash
$ nix-shell -p python3Packages.tensorflowWithCuda
~~~

and I am in a shell with CUDA Tensorflow.

At any rate, I'll try to address your other comment and then merge this change.

sticker-graph/sticker_graph/__init__.py
danieldk: Done!

PR #51, opened 2019-05-18
Adds the transformer model.

twuebi:
A good chunk of these lines are docs ;)
___________________________
#### This PR adds the transformer model. 

In detail:
* the feed-forward residual layers
* the encoder assembly 
* the write script including commandline args
________________________
I did some preliminary experiments for topo-field tagging and dependency parsing, both perform reasonably well.

### DIFF
--- a/sticker-graph/sticker-write-transformer-graph
+++ b/sticker-graph/sticker-write-transformer-graph
@@ -1,0 +1,116 @@
+#!/usr/bin/env python3
+
+import toml
+
+from sticker_graph.transformer_model import TransformerModel
+from sticker_graph.write_helper  import get_common_parser, parse_common_config, create_graph
+
+
+if __name__ == '__main__':
+    parser = get_common_parser()
+
+    parser.add_argument(
+        "--num_layers",
+        help="Number of transformer layers.",
+        type=int,
+        default=5)
+    parser.add_argument(
+        "--num_heads",
+        metavar="NUM_HEADS",
+        help="Number of attention heads.",
+        type=int,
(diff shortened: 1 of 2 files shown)


danieldk:
Looks good! Thanks for the well-documented code, it made it substantially easier to review the code. I would like to do some minor testing before merging.

danieldk:
And `apt-get upgrade` broke the whole CUDA installation. Now I have fixed everything (pinned CUDA to a specific version), but both GPUs are used on `turing-sfb`, so this has to wait until there is a GPU available again.

sebpuetz:
I'll stop one of the models. 

Did the upgrade happen on turing? I wanted to run some experiments earlier and from one moment to the other I got a message that libcublas.so.9.2 was missing.

sebpuetz:
GPU1 is free now.

danieldk:
Thanks! Shouldn't take too long, just want to do a few epochs to check if everything is working.

Indeed, this was on turing. The update bumped CUDA to 10.1 (not supported by Tensorflow), I set the CUDA symlink to a parallel installed 10.0, but the kernel driver and the CUDA libraries were out of sync, and the correct driver for 10.0 was not in the repo anymore. I now removed the NVIDIA CUDA repo and installed a specific version (so future updates shouldn't update CUDA at the same time).

danieldk:
Had to dial back the hyperparameters quite a bit for our Tesla M60 not to go out of memory, but seems to work great otherwise. Great work!

danieldk (approved):

PR #58, opened 2019-05-22
Configure learning rate schedule with sticker-train options

danieldk:
This moves the learning rate settings out of the configuration file.

This is also the last larger planned change before branching release-0.3.

### DIFF
--- a/sticker-utils/src/bin/sticker-train.rs
+++ b/sticker-utils/src/bin/sticker-train.rs
@@ -5,10 +5,11 @@ use std::io::BufReader;
 use clap::Arg;
 use failure::Error;
 use indicatif::ProgressStyle;
+use ordered_float::NotNan;
 use stdinout::OrExit;
 use sticker::depparse::{RelativePOSEncoder, RelativePositionEncoder};
 use sticker::tensorflow::{
-    ConllxDataSet, DataSet, LearningRateSchedule, TaggerGraph, TaggerTrainer,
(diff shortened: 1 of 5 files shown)


sticker-utils/src/bin/sticker-train.rs
```diff
@@ -5,29 +5,80 @@ use std::io::BufReader;
...
+
 impl TrainApp {
     fn new() -> Self {
         let matches = sticker_app("sticker-train")
+            .arg(
+                Arg::with_name(INITIAL_LR)
+                    .long("lr")
+                    .value_name("LR")
+                    .help("Initial learning rate")
+                    .default_value("0.01"),
```
sebpuetz, line 59: Iirc, 0.01 is a fairly high value, are you using that number in some models?

sebpuetz (approved):

sticker-utils/src/bin/sticker-train.rs
danieldk: It's my default for RNN models.

sebpuetz: I see, I was probably thinking about the clause models. I think i used 0.005 or 0.001 on those.

danieldk: The convolution models typically even diverge with 0.01, but I guess RNNs are currently the best default option. It's definitely something we should add to the documentation.

PR #60, opened 2019-05-22
Deny unknown fields in configuration files

danieldk:
My config files have accumulated quite some unused configuration options since the beginning of toponn/sticker. Besides failing on missing options, we should also fail on non-existing options. Keeps all the configuration files clean ;).


sebpuetz (approved):
Massive changes, whole 5 added attributes!

danieldk:
> Massive changes, whole 5 added attributes!

Yeah, serde is awesome!

PR #62, opened 2019-05-22
Add support for continuing training

danieldk:
This PR bundles two changes:

1. `TaggerTrainer` is separated into a separate module `trainer`.
2. Support for continuing training is added to `TaggerTrainer` and `sticker-train`.

Tested continuing a parser model, seems to work great!

Travis-CI seems to be down right now, but a build is running on sr.ht:

https://builds.sr.ht/~danieldk/job/63746

### DIFF
--- a/sticker-utils/src/bin/sticker-train.rs
+++ b/sticker-utils/src/bin/sticker-train.rs
@@ -3,7 +3,7 @@ use std::hash::Hash;
 use std::io::BufReader;
 
 use clap::Arg;
-use failure::Error;
+use failure::{Error, Fallible};
 use indicatif::ProgressStyle;
 use ordered_float::NotNan;
 use stdinout::OrExit;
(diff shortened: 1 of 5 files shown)


sticker-utils/src/bin/sticker-train.rs
```diff
@@ -51,6 +53,12 @@ impl TrainApp {
 impl TrainApp {
     fn new() -> Self {
         let matches = sticker_app("sticker-train")
+            .arg(
+                Arg::with_name(CONTINUE)
+                    .long("continue")
+                    .value_name("PARAMS")
+                    .help("Continue training from a parameter file"),
```
sebpuetz: maybe change help message to "Continue training from parameter file specified in config"?

sebpuetz (approved):

sticker-utils/src/bin/sticker-train.rs
danieldk: But it is not continuing from the parameter file specified in config ;). It uses the parameter file that is given as an argument to `--continue`.

sebpuetz: Just assumed this because I didn't recognize that `.value_name()` implicitly sets `.takes_value(true)`, but I guess that should be expected since a "value_name" doesn't make sense without an argument.

sebpuetz: Might still make sense to add some info like appending "e.g. epoch-50" to the help message since each epoch produces two files and it might be unclear to users what the "parameter file" is

danieldk: > Just assumed this because I didn't recognize that `.value_name()` implicitly sets `.takes_value(true)`, but I guess that should be expected since a "value_name" doesn't make sense without an argument.

I will add it for explicitness. The clap docs do it as well.

PR #63, opened 2019-05-23
Add the --save-batch option that saves after a given number of batches

danieldk:


sticker-utils/src/bin/sticker-train.rs
```diff
@@ -22,9 +22,71 @@ static LR_SCALE: &str = "LR_SCALE";
...
+impl From<SaveSchedule> for SaveScheduler {
+    fn from(schedule: SaveSchedule) -> Self {
+        SaveScheduler {
+            batch: 0,
+            epoch: 0,
+            schedule,
+        }
+    }
+}
+
```
sebpuetz, line 93: Maybe reorder the methods and structs in this file such that main is either at the top or the bottom of the file?

sebpuetz (approved):
This looks fine, although I'm unsure if saving after each epoch shouldn't be default and saving after every `n` batches should be an additional option.

I guess it's possible to figure out where an epoch ended, but depending on the `n` we might not get the actual end of an epoch.

If you don't like the idea, maybe add the epoch to the filename and cycle the batches? E.g. `epoch-i-batch-j` for the `j`th batch in the `i`th epoch?

danieldk:
> This looks fine, although I'm unsure if saving after each epoch shouldn't be default and saving after every `n` batches should be an additional option.

It's like that in this PR. Default: save after each epoch, use `--save-batch` and it will switch to saving after the given number of batches.

Or do you mean that both should be done --- always save after an epoch, and in addition after N batches?

> If you don't like the idea, maybe add the epoch to the filename and cycle the batches? E.g. `epoch-i-batch-j` for the `j`th batch in the `i`th epoch?

I have to think a bit about this. I think the only situation you'd want to use batch-based saving is during pretraining. However, in pretraining you probably don't care much about epochs, since you presumably do not have validation data anyway. But it does help to keep the saved batch numbers down when pretraining does several iterations over the data.

sticker-utils/src/bin/sticker-train.rs
danieldk: Good idea!

danieldk: I guess the ordering of `main` in the middle was already started by previous commits. So I'll do a separate clean up PR.

sebpuetz:
> Or do you mean that both should be done --- always save after an epoch, and in addition after N batches?

That's what I meant, default save after each epoch and in addition to that optionally save after ever N batches.

> > If you don't like the idea, maybe add the epoch to the filename and cycle the batches? E.g. `epoch-i-batch-j` for the `j`th batch in the `i`th epoch?
> 
> I have to think a bit about this. I think the only situation you'd want to use batch-based saving is during pretraining. However, in pretraining you probably don't care much about epochs, since you presumably do not have validation data anyway. But it does help to keep the saved batch numbers down when pretraining does several iterations over the data.

I didn't have any particular use-case in mind, I was just thinking that it might be informative to know whether the model had completed a full cycle through the data or not at a specific batch number.

danieldk:
> I didn't have any particular use-case in mind, I was just thinking that it might be informative to know whether the model had completed a full cycle through the data or not at a specific batch number.

That would indeed be informative, I'll add that.

PR #67, opened 2019-05-26
Adds the summary writer in write_helper.

twuebi:
This is the first step towards integrating tensorboard summaries.

The `_create_file_writer_generic_type` method is necessary to define `logdir` per run, in contrast to per graph definition. This is useful when experimenting with learning rates or batch sizes. 

This PR should not have an impact besides adding the ops to the graph. The next PRs will add the actual summary ops and the necessary bits in `sticker-train`.

### DIFF
--- a/sticker-graph/sticker_graph/write_helper.py
+++ b/sticker-graph/sticker_graph/write_helper.py
@@ -11,6 +11,35 @@ def read_shapes(args):
     return shapes
 
 
+def _create_file_writer_generic_type(logdir,
+                                    name="logdir",
+                                    max_queue=None,
+                                    flush_millis=None,
+                                    filename_suffix=None):
+    """
+    This method mirrors `tensorflow.contrib.summary.create_file_writer`. Unlike
+    `summary.create_file_writer`, this method accepts a placeholder as `logdir`.
+    """
+    from tensorflow.python.ops.summary_ops_v2 import _make_summary_writer
+    from tensorflow.python.ops.gen_summary_ops import create_summary_file_writer
+
+    logdir = tf.convert_to_tensor(logdir)
+    with tf.device("cpu:0"):
+        if max_queue is None:
+            max_queue = tf.constant(10)
+        if flush_millis is None:
+            flush_millis = tf.constant(2 * 60 * 1000)
+        if filename_suffix is None:
+            filename_suffix = tf.constant(".v2")
+        return _make_summary_writer(
+            name,
+            create_summary_file_writer,
+            logdir=logdir,
+            max_queue=max_queue,
+            flush_millis=flush_millis,
+            filename_suffix=filename_suffix)
+
+
 def create_graph(config, model, args):
     shapes = read_shapes(args)
     graph_filename = args.output_graph_file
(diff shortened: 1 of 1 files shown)


twuebi:
ci failure seems to be unrelated to this PR. 

```
error: identical conversion
  --> sticker/src/collector.rs:59:25
   |
59 |         for encoding in self.encoder.encode(sentence)? {
   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider removing `self.encoder.encode(sentence)?()`: `self.encoder.encode(sentence)?`
```

It seems to be a recent addition, probably rust 1.35.

danieldk:
> It seems to be a recent addition, probably rust 1.35.

This is a bug in clippy, desugaring `?` adds an `into()`, which is indeed superfluous in this case, but nothing that can be improved in code.

https://github.com/rust-lang/rust-clippy/issues/3944
https://github.com/rust-lang/rust-clippy/issues/3913

I'll temporarily use `1.34.2` in place of stable and retrigger the build.

sticker-graph/sticker_graph/write_helper.py
```diff
@@ -19,18 +48,27 @@ def create_graph(config, model, args):
     tfconfig = tf.ConfigProto(gpu_options=gpuopts)
 
     with tf.Graph().as_default(), tf.Session(config=tfconfig) as session:
-        with tf.variable_scope("model", reuse=None):
-            model(config=config, shapes=shapes)
+        logdir = tf.placeholder(shape=[], name="logdir", dtype=tf.string)
+        summary_writer = _create_file_writer_generic_type(logdir)
+
+        with summary_writer.as_default(), tf.contrib.summary.always_record_summaries():
+            with tf.variable_scope("model", reuse=None):
```
danieldk, line 55: Just to be sure: does this `variable_scope` need to be scoped inside the summary_writer? Or could it be added to the conjunction above? (If so, that would be clearer.)

sticker-graph/sticker_graph/write_helper.py
```diff
@@ -19,18 +48,27 @@ def create_graph(config, model, args):
...
+                model(config=config, shapes=shapes)
 
-        tf.variables_initializer(tf.global_variables(), name='init')
+            tf.group(tf.contrib.summary.summary_writer_initializer_op(),
+                     name="summary_init")
+            tf.variables_initializer(tf.global_variables(), name='init')
```
danieldk, line 60: Same question as above, does this need to be in that scope?

danieldk (commented):
Just two small questions about scoping.

sticker-graph/sticker_graph/write_helper.py
twuebi: Something like that?

```py
with summary_writer.as_default(), tf.contrib.summary.always_record_summaries(), tf.variable_scope("model",
                                                                                                  reuse=None):
    model(config=config, shapes=shapes)
    tf.group(tf.contrib.summary.summary_writer_initializer_op(),
             name="summary_init")
    serialized_graph = session.graph_def.SerializeToString()
    serialized_graph_tensor = tf.convert_to_tensor(serialized_graph)
    tf.contrib.summary.graph(serialized_graph_tensor, 0, name='graph_write')
```

It should work. I didn't go for it since it would place the summary ops inside the `model` scope. Having the 3 contexts in the same `with` also seems visually unclear. I like the original version as it seperates the summary stuff from the model related scope.

sticker-graph/sticker_graph/write_helper.py
twuebi: It doesn't need to. I kept it so that the `graph_write` op will capture the full graph, including the init, save and restore ops.

sticker-graph/sticker_graph/write_helper.py
danieldk: Like that yes. Well, with these things I always worry that Tensorflow context managers introduce new variable scopes and the ops are renamed ;)

I see your point about summary op scopes. So, then a slightly different question: does the model variable scope need to be within or together with `summary_writer` at all? Can they be placed in parallel?

~~~python
with tf.variable_scope("model", reuse=None):
  # Model stuff

with summary_writer.as_default(), tf.contrib.summary.always_record_summaries():
 # Summary stuff
~~~

Or do the ops have to be created within the `summary_writer` context manager for it to pick up the graph nodes?

twuebi: I just tested it, it doesn't pick up the summary ops if the model is not created inside the scope. I did not experience any context manager related issues. I'd not expect these managers to mess with any variable scope.

danieldk (approved):
Ok, then this is good to go!

danieldk:
Oh, could you rebase this on master? Want to avoid merge commits (which GitHub wants to generate).

twuebi:
Sure

PR #72, opened 2019-06-07
Adds Tensorboard summaries.

twuebi:
This PR adds a parameter --logdir to sticker-train.

If it is provided, summaries are written to that directory. The summaries entail accuracy, loss, learning rate, and if provided gradient norms.

For the RNN-model, no gradient norms are logged because no gradient clipping is performed. This may be something for later

Screenshot from some transformer training:

tensorboard
_______________________________________________________

The PR got a bit bulkier than planned. Originally I had planned to first add the `init_logdir` method and the `--logdir` parameter and then in a separate PR the accuracy, loss, etc summaries, but for some reason, the graph is not written if no other summaries are added. If wanted, I can still split this PR into two, one for graph writing, the other to add the actual summaries. This one could then serve as a functional basis to see what the result would look like.

### DIFF
--- a/sticker-graph/sticker_graph/conv_model.py
+++ b/sticker-graph/sticker_graph/conv_model.py
@@ -194,13 +194,17 @@ class ConvModel(Model):
             predictions = self.predictions("tag", logits)
             self.top_k_predictions("tag", logits, config.top_k)
 
-        self.accuracy("tag", predictions, self.tags)
+        acc = self.accuracy("tag", predictions, self.tags)
 
         # Optimization with gradient clipping. Consider making the gradient
         # norm a placeholder as well.
         lr = tf.placeholder(tf.float32, [], "lr")
         optimizer = tf.train.AdamOptimizer(lr)
         gradients, variables = zip(*optimizer.compute_gradients(loss))
-        gradients, _ = tf.clip_by_global_norm(gradients, 1.0)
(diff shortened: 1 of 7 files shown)


danieldk (approved):
Looks good to me!

Note that this breaks backwards compatibility with existing graphs. I think the next release should be the last release that breaks graph compatibility for a while, since breaking changes are annoying to users. (I know of at least one company that is currently evaluating sticker.)

But since this is a very useful feature, I think we should do this before freezing things for a while.

twuebi:
> Note that this breaks backwards compatibility with existing graphs. I think the next release should be the last release that breaks graph compatibility for a while, since breaking changes are annoying to users. (I know of at least one company that is currently evaluating sticker.)

That's true, but it's really only the graph, the parameters can still be used if the graph is rewritten. For full compatibility, we could wrap the tensorboard ops in the rust graph in options or manually add no-ops from rust, while disabling the logging. That'd be weight to carry around for all future versions, so I guess breaking the compatibility once with the graph rewrite option seems the better way here.

danieldk:
> but it's really only the graph, the parameters can still be used if the graph is rewritten

I know, but these are not things that you want to explain downstream users. They should be able to upgrade sticker without having to fix their models. We are still at 0.3, so I think it is ok to do it properly now, but I would really like to move to strong backwards compat.

If you break it once or twice for downstream users, they will just stop using it. Or stick with one version for eternity and miss out on bug fixes or optimizations.

PR #75, opened 2019-06-11
Add sticker-pretrain

danieldk:
This change splits out support for pretraining into a separate
utility.

### DIFF
--- a/sticker-utils/src/bin/sticker-pretrain.rs
+++ b/sticker-utils/src/bin/sticker-pretrain.rs
@@ -1,6 +1,6 @@
 use std::fs::File;
 use std::hash::Hash;
-use std::io::BufReader;
+use std::io::{BufReader, Seek, SeekFrom};
 
 use clap::Arg;
 use failure::{Error, Fallible};
@@ -8,9 +8,7 @@ use indicatif::ProgressStyle;
 use ordered_float::NotNan;
 use stdinout::OrExit;
 use sticker::depparse::{RelativePOSEncoder, RelativePositionEncoder};
-use sticker::tensorflow::{
-    ConllxDataSet, DataSet, LearningRateSchedule, PlateauLearningRate, TaggerGraph, TaggerTrainer,
-};
+use sticker::tensorflow::{ConllxDataSet, DataSet, TaggerGraph, TaggerTrainer};
 use sticker::{CategoricalEncoder, LayerEncoder, Numberer, SentVectorizer, SentenceEncoder};
 use sticker_utils::{
     sticker_app, CborRead, CompletedUnit, Config, EncoderType, LabelerType, ReadProgress,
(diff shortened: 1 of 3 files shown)


sticker-utils/src/bin/sticker-pretrain.rs
```diff
@@ -0,0 +1,346 @@
...
+        eprintln!(
+            "Epoch {} (validation): loss: {:.4}, acc: {:.4}, best epoch: {}, best acc: {:.4} {}",
+            epoch, loss, acc, best_epoch, best_acc, epoch_status
+        );
+    }
+
+    Ok(())
+}
+
+#[allow(clippy::too_many_arguments)]
```
sebpuetz, line 271: Large lists of arguments are something that we see quite often in our tools. Maybe defining a struct that holds the relevant parameters could simplify some of the binaries?

sticker-utils/src/save.rs
```diff
@@ -0,0 +1,90 @@
...
+        self.epoch
+    }
+
+    /// Save the model parameters when a save point has been reached.
+    pub fn save(&mut self, trainer: &TaggerTrainer, completed: CompletedUnit) -> Fallible<()> {
+        match completed {
+            CompletedUnit::Epoch => {
+                match self.schedule {
+                    SaveSchedule::Epoch | SaveSchedule::EpochAndBatches(_) => {
+                        trainer.save(format!("pretrain-epoch-{}", self.epoch))?
```
sebpuetz: Should the prefix `pretrain` be an argument? As far as I can tell, this is also the prefix for saved epochs when using `sticker-train`.

sebpuetz (changes requested):

sticker-utils/src/save.rs
danieldk: Good catch! will fix that tomorrow!

danieldk: Fixed.

sticker-utils/src/bin/sticker-pretrain.rs
danieldk: I could do that in a separate PR, though I want to investigate first how ad-hoc the structs would be.

sebpuetz: Wrapping all the parameters in a struct and passing that wouldn't really alleviate anything, maybe it's possible to implement `run_epoch` on a struct. But that's nothing for this PR, i just saw the line while going through the changed files.

danieldk: Yeah, I already felt bad when I disabled that clippy lint ;).

danieldk:
Ready to go?

sebpuetz:
Didn't catch anything else when I went through the changes in the morning

PR #78, opened 2019-06-16
Update license to the Blue Oak Model License 1.0.0

danieldk:
I also added some wording to CONTRIBUTORS that allows us to upgrade to a
newer version of the license when necessary.

Acked by Tobias Pütz.


CONTRIBUTORS
```diff
@@ -0,0 +1,11 @@
...
+
+sticker is provided to you by its contributors under the Blue Oak Model License
+Version 1.0.0 (see LICENSE.md). Both Daniël de Kok and Tobias Pütz may
+individually, at their own discretion, relicense future versions of sticker
+under any later version of the Blue Oak Model License.
+
+Contributors:
+
+Daniël de Kok <me@danieldk.eu>
+Tobias Pütz <tobias.puetz@student.uni-tuebingen.de>
```
twuebi: Perhaps change to tobias.puetz@uni-tuebingen.de which may be the more permanent address.

twuebi (approved):

CONTRIBUTORS
danieldk: Good idea! Fixed.

PR #79, opened 2019-06-17
sticker-write-conv-graph: bring back hidden_size option

danieldk:


twuebi:
Oh, this shouldn't have happened. We should move away from the redundant `DefaultConfig` + `argparse` setup.

twuebi (approved):

PR #80, opened 2019-06-18
Purge DefaultConfig from sticker-graph

danieldk:
I replaced `config` by `args` throughout, since that is now our defacto configuration. `args` also guarantees that whatever member of `args` is used is also defined in the option parser, which should avoid bugs like the one we had yesterday.


twuebi (approved):
looks good to me.

Unrelated to this PR but caught my eye:

maybe print model information for all three models the same way. Currently, the transformer script prints toml, rnn and conv use a format string. We could move the info print to `create_graph`.

danieldk:
> looks good to me.
> 
> Unrelated to this PR but caught my eye:
> 
> maybe print model information for all three models the same way. Currently, the transformer script prints toml, rnn and conv use a format string. We could move the info print to `create_graph`.

Great idea, especially because the RNN/conv information is now also incomplete.

PR #83, opened 2019-06-18
Add option for residual connections to the rnn.

twuebi:
Adds `--residual` to `sticker-write-rnn-graph` which enables residual skip connections of the
rnn layers.

### DIFF
--- a/sticker-graph/sticker-write-rnn-graph
+++ b/sticker-graph/sticker-write-rnn-graph
@@ -30,7 +30,12 @@ if __name__ == '__main__':
         type=float,
         help="input keep probability",
         default=0.80)
-
+    parser.add_argument(
+        "--residual",
+        action='store_true',
+        default=False,
+        help="Use residual skip connections between RNN layers."
+    )
     args = parser.parse_args()
 
     model = RNNModel
--- a/sticker-graph/sticker_graph/rnn_model.py
+++ b/sticker-graph/sticker_graph/rnn_model.py
@@ -1,7 +1,8 @@
 import tensorflow as tf
 from tensorflow.contrib.layers import batch_norm
-from sticker_graph.model import Model
 
+from sticker_graph.model import Model
+import sticker_graph.vendored
 
 def dropout_wrapper(
         cell,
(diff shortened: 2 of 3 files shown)


sticker-graph/sticker-write-rnn-graph
```diff
@@ -30,7 +30,13 @@ if __name__ == '__main__':
         type=float,
         help="input keep probability",
         default=0.80)
-
+    parser.add_argument(
+        "--residual",
+        type=bool,
+        action='store_true',
+        default=False,
+        help="Use residual skip connections between rnn layers."
```
danieldk: Nitpicking: s/rnn/RNN/

sticker-graph/sticker_graph/rnn_model.py
```diff
@@ -22,6 +22,79 @@ def dropout_wrapper(
         state_keep_prob=state_keep_prob)
 
 
+def stack_bidirectional_dynamic_rnn(cells_fw,
```
danieldk: We should put a comment up here with the source of the original function. It would be nicer if it was in a separate file (maybe `vendored.py` ?), because then we could also include the copyright header from:

https://github.com/tensorflow/tensorflow/blob/r1.13/tensorflow/contrib/rnn/python/ops/rnn.py

It should probably also have a 1-2 line description of what was modified compared to the upstream function.

danieldk (changes requested):
Some small doc/copyright requests.

sticker-graph/sticker_graph/rnn_model.py
twuebi: The same thing should also happen in `write_helper.py` for `_create_file_writer_generic_type`. Both things in one go in another PR, or `stack_bidirectional_rnn` now and then in another PR `_create_file_writer_generic_type`?

sticker-graph/sticker-write-rnn-graph
twuebi: same thing in `--hidden_size`

sticker-graph/sticker_graph/rnn_model.py
danieldk: The `_create_file_writer_generic_type` change should go in a separate PR.

sticker-graph/sticker-write-rnn-graph
danieldk: Oops :)

danieldk (approved):

danieldk:
Had to be rebased, so merged into `master` by hand.

PR #87, opened 2019-06-23
Fix Tensorboard for TF 1.14.

### LINKED ISSUE #86: Tensorboard broken with tf 1.14
https://github.com/tensorflow/tensorflow/commit/826027dbd4277a2636fc2935ed245700fd01e7cd#diff-335a6620279c6fa5d7f70c37c91265a3

this broke our `_create_file_writer_generic_type` in `vendored.py`, it was introduced with Tensorflow 1.14. I'm looking into it.

twuebi:
fixes #86 

Use `ResourceSummaryWriter` for Tensorflow >= 1.14 and `_make_summary_writer` for Tensorflow < 1.14. Currently, writing graphs on 1.14 fails as ` _make_summary_writer` was removed in 1.14.

### DIFF
--- a/sticker-graph/sticker_graph/vendored.py
+++ b/sticker-graph/sticker_graph/vendored.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # ==============================================================================
+import functools
+
 import tensorflow as tf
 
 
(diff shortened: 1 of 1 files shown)


sticker-graph/sticker_graph/vendored.py
```diff
@@ -114,10 +115,23 @@ def _create_file_writer_generic_type(logdir,
...
+                logdir=logdir,
+                max_queue=max_queue,
+                flush_millis=flush_millis,
+                filename_suffix=filename_suffix))
+    else:
+        from tensorflow.python.ops.summary_ops_v2 import _make_summary_writer
         return _make_summary_writer(
-            name,
-            create_summary_file_writer,
+             name,
```
danieldk: Are these extra spaces intentional? It's hard to see in GitHub, but it seems that as a result they are not aligned with the other arguments anymore? (Again, GitHub may be messing up the layout here.)

twuebi: You're right. Since tensorflow-rocm is not on 1.14 yet, I fixed it remotely using vim and didn't see that.

Should be fine now.

danieldk (approved):

PR #88, opened 2019-06-27
sticker-{pretrain,train}: introduce maxlen option

danieldk:
This argument skips sentences that are longer than the given length
during training.

Tested using an excessive model for a K20c, flies with `--maxlength 10` ;).


twuebi (approved):
looks good to me

PR #89, opened 2019-06-27
CI: Test graph writing scripts

danieldk:


danieldk:
@twuebi let me know if we should add some parameters to the tranformer model that we always want to test to work.

twuebi:
~~~sh
sticker-write-transformer-graph ../shapes g --outer_hsize 350 --pass_inputs --num_heads 7
sticker-write-transformer-graph ../shapes g --outer_hsize 350 --pass_inputs --num_heads 7 --embed_time --max_time 150
~~~

Either the second or both of them would be good #91

danieldk:
Ok, added these tests and rebased on #91.

twuebi:
Gah, forgot the most common use-case:

~~~sh
sticker-write-transformer-graph shapes trans-default.graph
~~~

danieldk:
Added

twuebi (approved):

PR #98, opened 2019-07-14
Restore compatibility with sticker 0.4 models

danieldk:
This change restores compatibility with models that are trained using
sticker 0.4. The only graph incompatibility is the addition of
training and validation summaries for TensorBoard. Make these ops
optional, so that prediction (and training) using old graphs continues
to work.

The ops could be made mandatory in some future version when the
sticker 0.4 models have all been replaced.


sticker/src/tensorflow/tagger.rs
```diff
@@ -150,8 +150,8 @@ impl TaggerGraph {
         let logdir_op = Self::add_op(&graph, op_names::LOGDIR_OP)?;
```
twuebi: If I'm not mistaken, neither, `logdir_op`, `graph_write_op` nor `summary_init_op` should be present in 0.4 graphs.

twuebi (commented):
Could you provide a 0.4 graph for testing? I got the suspicion that we either crash because `graph_write_op`, `logdir_op` and `summary_init_op` are not in the 0.4 graphs or that we get an unhandled Tensorflow error when providing the `--logdir` option in conjunction with a 0.4 graph. The calls to `init_logdir` in `sticker-(pre)train` depend on `--logdir` and will attempt to access these ops:

~~~rust
    pub(crate) graph_write_op: Operation,
    pub(crate) logdir_op: Operation,
    pub(crate) summary_init_op: Operation,
~~~


Maybe extend the CI with graph loading to be on the safe side?

sticker/src/tensorflow/tagger.rs
danieldk: Thanks! Indeed, this should be an `Option` as well.

danieldk:
> Could you provide a 0.4 graph for testing? I got the suspicion that we either crash because `graph_write_op`, `logdir_op` and `summary_init_op` are not in the 0.4 graphs or that we get an unhandled Tensorflow error when providing the `--logdir` option in conjunction with a 0.4 graph. The calls to `init_logdir` in `sticker-(pre)train` depend on `--logdir` and will attempt to access these ops:

I have only tested prediction since that was my primary goal. I'll test training as well. You can find a bunch of 0.4 models here:

https://blob.danieldk.eu/sticker-models/

> Maybe extend the CI with graph loading to be on the safe side?

That's a great idea. I do this in dpar and it has caught some bugs. Would be nice to have a graph per version, so that we can see how far we can go back.

The other thing I am sometimes worried about is that *with different Tensorflow versions* the graph loads, but that there is somehow a mismatch between the graph in Tensorflow itself and the parameters, such that the parameters load, but the output is garbage. I haven't observed this in practice yet though, and I have sometimes used graphs + parameters from several versions back.

Though with distributed models, I am wondering if I shouldn't stick with the original version that the model was trained with. The Tensorflow dynamic library should work for an eternity and it is easy to have several version in parallel with Nix (which I am building the Docker images with).

sticker/src/tensorflow/tagger.rs
danieldk: But it is indeed surprising that it loaded the graph at all. It could have been that the graphs were created using a version between 0.4 and `HEAD`. I will create a graph with pure 0.4.0 and add that as a test.

twuebi:
> That's a great idea. I do this in dpar and it has caught some bugs. Would be nice to have a graph per version, so that we can see how far we can go back.

Having graphs per version sounds nice, I'd prefer it over the package-with-back-then-TF-version.

> The other thing I am sometimes worried about is that _with different Tensorflow versions_ the graph loads, but that there is somehow a mismatch between the graph in Tensorflow itself and the parameters, such that the parameters load, but the output is garbage. I haven't observed this in practice yet though, and I have sometimes used graphs + parameters from several versions back.

If that should ever be the case, I'd hope that the loading op crashes. 

> Though with distributed models, I am wondering if I shouldn't stick with the original version that the model was trained with. The Tensorflow dynamic library should work for an eternity and it is easy to have several version in parallel with Nix (which I am building the Docker images with).

Trade-off between potential optimizations and stability? I'd expect a library to be very clear (at least deprecation warning or plain crash) if such breaking changes were going to be introduced.

danieldk:
> > The other thing I am sometimes worried about is that _with different Tensorflow versions_ the graph loads, but that there is somehow a mismatch between the graph in Tensorflow itself and the parameters, such that the parameters load, but the output is garbage. I haven't observed this in practice yet though, and I have sometimes used graphs + parameters from several versions back.
> 
> If that should ever be the case, I'd hope that the loading op crashes.
> 
> [...]
> 
> Trade-off between potential optimizations and stability? I'd expect a library to be very clear (at least deprecation warning or plain crash) if such breaking changes were going to be introduced.

They do guarantee backwards compat for parameters and graphs:

> Many TensorFlow users save graphs and trained models to disk for later evaluation or additional training, but end up running their saved graphs or models on a later release. In compliance with semver, any graph or checkpoint written out with one version of TensorFlow can be loaded and evaluated with a later version of TensorFlow with the same major release. However, we will endeavor to preserve backwards compatibility even across major releases when possible, so that the serialized files are usable over long periods of time.

Source: https://www.tensorflow.org/guide/version_compat#compatibility_of_graphs_and_checkpoints

So at least with any version before 2.0 we shouldn't expect problems. Let's hope for the best ;). Once we switch to 2.0, we should probably stick with 1.14 for old models.

danieldk:
Ok, the other ops are covered now as well, and I added unit tests for loading 0.4 conv/rnn/transformer graphs.

sticker/src/tensorflow/tagger.rs
```diff
@@ -146,12 +146,12 @@ impl TaggerGraph {
 
         let train_op = Self::add_op(&graph, op_names::TRAIN_OP)?;
 
-        let graph_write_op = Self::add_op(&graph, op_names::WRITE_GRAPH_OP)?;
-        let logdir_op = Self::add_op(&graph, op_names::LOGDIR_OP)?;
-        let summary_init_op = Self::add_op(&graph, op_names::SUMMARY_INIT_OP)?;
+        let graph_write_op = Self::add_op(&graph, op_names::WRITE_GRAPH_OP).ok();
```
twuebi, line 149: I am not entirely happy about quietly handling all errors in these `add_op` calls as _just-do-not-write-summaries_.  E.g. if a name mismatch is introduced somewhere in python land, we're treating it the same way as a regular 0.4 graph. This would then eventually crash at the `init_logdir` call in `sticker-(pre)train`.

A more verbose option:

~~~rust
let graph_write_op = match Self::add_op(&graph, op_names::WRITE_GRAPH_OP) {
    Ok(op) => op,
    Err(_) => {
        eprintln!("Couldn't find {} in the graph. Summary writing is not supported!\
         This is likely due to a graph from sticker <0.5", op_names::WRITE_GRAPH_OP);
        None
    }
};
~~~

I am also not a fan of that option, since it throws a warning message meant for users who actually want to train at everyone, including probably uninterested tagging-users.

twuebi (approved):
Seems fine. 

I am a bit uncertain whether we want to handle all errors with finding the summary ops quietly. If it's just meant as an intermediate solution until we migrated all models to >0.4, it's probably not an issue.

sticker/src/tensorflow/tagger.rs
danieldk: Indeed, we should avoid such messages for users of 0.4 models. There are two other obvious options:

1. Store `Result` rather than an `Option`, so that we still carry the actual errors in case someone tries to use the ops.
2. Use lazy initialization, so that the op is only looked up (once) on use. The overhead should be small, because after initialization only an atomic is used to check whether the value is initialized (a mutex is only used during initialization). It is questionable when the benefit is over (1), except that the lookup is really only done when necessary and it feels semantically correct.

Of course, in both cases, the error would still be in `init_logdir` (which does not feel wrong to me, because at that point you are doing something that is incompatible with the graph). But at least, you could print the actual error messages.

danieldk: Third option: make a new constructor `new_with_logdir` that is used when the option is passed to `sticker-train`. This would have a small benefit that we could fail in the constructor. Probably makes the constructor even more expensive though.

twuebi: My bad, I didn't realize that we actually have verbose messages in `init_logdir`:

~~~rust
args.add_target(
        self.graph
            .summary_init_op
            .as_ref()
            .ok_or_else(|| err_msg("Graph does not support writing of summaries"))?,
);
~~~

This is the only place where we can expect issues and we are verbose about it, so I think your PR as is should be fine.

danieldk: Ack! Thanks for the review!

PR #100, opened 2019-07-15
Fix CRF support

### LINKED ISSUE #99: Training with CRFs fails
When setting --crf on write-graph-rnn, graphs can't be loaded. Seems like `tag_top_k_probs` is not created in that case.

```
python sticker-write-rnn-graph shapes graph --hidden_size 100 --crf
sticker-train sticker.conf ../large-train-nongold-pos.ner ../dev-nongold-pos.ner
Cannot construct trainer: Unavailable: Operation "model/tag_top_k_probs" not found
```

danieldk:
No op named 'tag_top_k_probs' was created when '--crf' was used. This change
adds this op to graphs with a CRF layer.

Since CRF decoding does not provide per-tag probabilities, use a probability
of 1 for each rag.

Fixes #99.

### DIFF
--- a/sticker-graph/sticker_graph/model.py
+++ b/sticker-graph/sticker_graph/model.py
@@ -57,6 +57,11 @@ class Model:
         predictions, _ = tf.contrib.crf.crf_decode(
             logits, transitions, self.seq_lens)
         top_k_predictions = tf.expand_dims(predictions, 2)
+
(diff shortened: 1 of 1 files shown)


sebpuetz (approved):
Looks fine, although I didn't find time to run it yet.

PR #107, opened 2019-07-16
Store graph construction args to the graph

danieldk:
Also provide the sticker-graph-metadata utility to print the args.

### DIFF
--- a/sticker-graph/sticker_graph/write_helper.py
+++ b/sticker-graph/sticker_graph/write_helper.py
@@ -13,7 +13,6 @@ def read_shapes(args):
 
 
 def create_graph(model, args):
-    write_args(model, args)
 
     shapes = read_shapes(args)
     graph_filename = args.output_graph_file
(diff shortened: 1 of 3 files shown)


sticker-graph/sticker_graph/write_helper.py
```diff
@@ -1,4 +1,5 @@
 import argparse
+import os
```
twuebi: Stray import?

twuebi (changes requested):

sticker-graph/sticker_graph/write_helper.py
danieldk: Indeed. I first added the program name to the metadata as well (needed `basename` for that), but then decided that having the model name in the metadata is probably enough.

danieldk: Fixed.

twuebi (approved):

PR #109, opened 2019-07-20
Restructure encoder modules

danieldk:
- The top-level depparse module is renamed to encoder::deprel.
- The Layer{Encoder,Decoder} are moved to the encoder::layer module.

Mostly a boring reshuffling of modules. But the structure should be a bit more logical as a result.


danieldk:
CI may fail on 1.31 because I changed quite a some imports, and the probably need `crate` prefix. Will fix them if CI indeed fails.

sebpuetz (approved):

PR #110, opened 2019-07-21
Improve encoder error handling

danieldk:


sebpuetz (approved):
Looks fine, although I think the string representation of the sentence or the debug representation of the token is more informative than just the form of the token missing something. That would allow to find the malformed line more easily without messing with the code.

danieldk:
> Looks fine, although I think the string representation of the sentence or the debug representation of the token is more informative than just the form of the token missing something. That would allow to find the malformed line more easily without messing with the code.

I agree. I think that including the whole sentence might be to long for an error. It seems that `Display` is not implemented for `Token` (AFAIR the reason is that the token does not have its dependency relations in the new `conllx` and does not know its position). `Debug` implementations are not so nice for output that a user may see. But I could make a `DisplayToken` wrapper for `Token` that at least outputs the form, lemma, cpos, and features (if present).

Edit: I wonder if multi-line errors are a thing, because then we could include the form/position and the full sentence (just its forms) on a separate line.

sebpuetz:
> Edit: I wonder if multi-line errors are a thing, because then we could include the form/position and the full sentence (just its forms) on a separate line.

I don't see a problem with reporting an error on multiple lines. It would allow to find the right sentence in a sensible manner and I don't see obvious downsides.

danieldk:
> > Edit: I wonder if multi-line errors are a thing, because then we could include the form/position and the full sentence (just its forms) on a separate line.
> 
> I don't see a problem with reporting an error on multiple lines. It would allow to find the right sentence in a sensible manner and I don't see obvious downsides.

I implemented this, but it is very repetitive. Then I realized it's probably the code that calls the encoder that should report the sentence in which the error occurred.

PR #111, opened 2019-07-30
Add support for RNN-based subword representations

danieldk:
When subword representations are enabled, sticker builds a string
tensor containing each form in the input. The forms are converted to
(truncated) byte representations in the Tensorflow graph. Each byte is
then mapped to a byte embedding. Subword representations are then
formed by applying a bidirectional RNN to the sequence of byte
embeddings.

In order to support the configuration of subword embeddings, the
sticker configuration format has been adjusted. An `input` section has
been added, which holds the new `subwords` boolean option. The
`embeddings` section is now a subsection of `input`. The use of
subword representations is communicated to the graph writers through
the shapes file.

---

Sorry for the large changeset! Adding subword representations touches
many parts of sticker. This should be the last *large* change before
sticker 1.0 and also the last time an incompatible change is made to the
configuration file format before 1.0.

### DIFF
--- a/sticker-graph/sticker_graph/model.py
+++ b/sticker-graph/sticker_graph/model.py
@@ -1,5 +1,7 @@
 import tensorflow as tf
 
+from sticker_graph.rnn import bidi_rnn_layers
+
 
 class Model:
     def __init__(self, args, shapes):
@@ -33,6 +35,7 @@ class Model:
             tf.nn.xw_plus_b(
                 x, w, b), [
                 batch_size, -1, n_outputs])
+
 
     def masked_softmax_loss(self, prefix, logits, labels, mask):
         # Compute losses
@@ -92,14 +95,9 @@ class Model:
 
         return probs, labels
 
-    def setup_placeholders(self):
-        self._is_training = tf.placeholder(tf.bool, [], "is_training")
-
-        self._tags = tf.placeholder(
-            tf.int32, name="tags", shape=[
-                None, None])
-
-        self._inputs = tf.placeholder(
+    def setup_inputs(self):
+        # Word/tag embeddings
+        inputs = tf.placeholder(
             tf.float32,
             shape=[
                 None,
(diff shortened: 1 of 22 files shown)


sticker-graph/sticker_graph/model.py
```diff
@@ -116,6 +139,55 @@ def setup_placeholders(self):
...
+        #
+        # Shape: [batch_size * seq_len, max_bytes_len, byte_embed_size]
+        bytes_shape = tf.shape(subword_bytes_padded)
+        byte_reprs = tf.reshape(
+            byte_reprs, [-1, bytes_shape[2], self.args.byte_embed_size])
+        byte_lens = tf.reshape(subword_lens, [-1])
+
+        with tf.variable_scope("byte_rnn"):
+            _, fw, bw = bidi_rnn_layers(self.is_training, byte_reprs, num_layers=self.args.subword_layers,
+                                        output_size=self.args.subword_hidden_size, output_keep_prob=self.args.subword_keep_prob, seq_lens=byte_lens)
```
twuebi: Maybe allow residual connections here?

twuebi: Possibly also the option to use GRU

danieldk: Great idea, I'll update the PR later.

danieldk: I have added options for both.

twuebi (approved):
Looks good, not entirely happy with calling it subwords, I'd rather go for byte-rnn or char-rnn. Although that wouldn't play nicely with adding char-convolutions later on.

danieldk:
> Looks good, not entirely happy with calling it subwords, I'd rather go for byte-rnn or char-rnn. Although that wouldn't play nicely with adding char-convolutions later on.

Couldn't come up with a better name yet.

BTW. this currently still has a bug (for non-latin1 unicode characters), since it uses the wrong decoding function. But I'll do a separate PR that fixes this.

PR #113, opened 2019-08-05
Fix byte encoding

danieldk:
tf.strings.unicode_decode decodes to unicode codepoints. So, any non-latin-1
character falls outside the range. This change introduces actual byte
encodings.

In some experiments, I found that with real byte representations, it helps
to have two layers in the byte RNN (presumably to combine characters with
multi-byte encodings), so this also changes the default hyperparameters.

### DIFF
--- a/sticker-graph/sticker_graph/model.py
+++ b/sticker-graph/sticker_graph/model.py
@@ -36,6 +36,22 @@ class Model:
                 x, w, b), [
                 batch_size, -1, n_outputs])
 
+    def decode_word(self, word, max_subword_len):
+        bytes = tf.decode_raw(word, tf.uint8)[:max_subword_len]
+        return tf.pad(bytes, [[0, max_subword_len - tf.shape(bytes)[0]]])
+
+    def decode_words(self, words, max_subword_len):
+        shape = tf.shape(words)
+
+        # Flatten, because map_fn only works on axis 0
+        flat = tf.reshape(words, [-1])
+
+        # Decode words.
+        decoded = tf.map_fn(lambda x: self.decode_word(x, max_subword_len), flat, dtype=tf.uint8)
+
+        # Restore shape
+        return tf.reshape(decoded, [shape[0], shape[1], max_subword_len])
+
 
     def masked_softmax_loss(self, prefix, logits, labels, mask):
         # Compute losses
@@ -143,19 +159,13 @@ class Model:
         # Convert strings to a byte tensor.
         #
         # Shape: [batch_size, seq_len, subword_len]
-        subword_bytes = tf.strings.unicode_decode(
-            self.subwords, input_encoding='UTF-8')
-        subword_bytes_padded = subword_bytes.to_tensor(
-            0)[:, :, :self.args.subword_len]
+        with tf.device("/cpu:0"):
+            subword_bytes = tf.cast(self.decode_words(self.subwords, self.args.subword_len), tf.int32)
 
-        # Get the lengths of the subwords. Only the last dimension should
-        # be ragged, so no actual padding should happen.
-        #
+        # Get the lengths of the subwords.
         # Shape: [batch_size, seq_len]
-        subword_lens = tf.math.minimum(
-            subword_bytes.row_lengths(
-                axis=-1).to_tensor(0),
-            self.args.subword_len)
+        with tf.device("/cpu:0"):
+            subword_lens = tf.count_nonzero(subword_bytes, axis=-1, dtype=tf.int32)
 
         # Lookup byte embeddings, this results in a tensor of shape.
         #
@@ -163,7 +173,7 @@ class Model:
         byte_embeds = tf.get_variable(
             "byte_embeds", [
                 256, self.args.byte_embed_size])
-        byte_reprs = tf.nn.embedding_lookup(byte_embeds, subword_bytes_padded)
+        byte_reprs = tf.nn.embedding_lookup(byte_embeds, subword_bytes)
 
         byte_reprs = tf.contrib.layers.dropout(
             byte_reprs,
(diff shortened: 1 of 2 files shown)


sticker-graph/sticker_graph/model.py
```diff
@@ -143,27 +159,25 @@ def subword_reprs(self):
         # Convert strings to a byte tensor.
         #
         # Shape: [batch_size, seq_len, subword_len]
-        subword_bytes = tf.strings.unicode_decode(
-            self.subwords, input_encoding='UTF-8')
-        subword_bytes_padded = subword_bytes.to_tensor(
-            0)[:, :, :self.args.subword_len]
+        with tf.device("/cpu:0"):
+            subword_bytes = tf.cast(self.decode_words(self.subwords, self.args.subword_len), tf.int32)
+        print(subword_bytes)
```
twuebi: Spurious print?

sticker-graph/sticker_graph/model.py
```diff
@@ -143,27 +159,25 @@ def subword_reprs(self):
...
         # be ragged, so no actual padding should happen.
         #
         # Shape: [batch_size, seq_len]
-        subword_lens = tf.math.minimum(
-            subword_bytes.row_lengths(
-                axis=-1).to_tensor(0),
-            self.args.subword_len)
+        with tf.device("/cpu:0"):
+            subword_lens = self.word_lens(self.subwords, self.args.subword_len)
+        print(subword_lens)
```
twuebi: Same as above?

sticker-graph/sticker_graph/model.py
danieldk: whoops, thanks for catching that! Should be fixed now.

sticker-graph/sticker_graph/model.py
```diff
@@ -188,6 +200,23 @@ def subword_reprs(self):
...
+        return tf.shape((tf.decode_raw(word, tf.uint8)[:max_subword_len]))[0]
+
+    def word_lens(self, words, max_subword_len):
+        shape = tf.shape(words)
+
+        # Flatten, because map_fn only works on axis 0
+        flat = tf.reshape(words, [-1])
+
+        # Decode words.
+        decoded = tf.map_fn(lambda x: self.word_len(x, max_subword_len), flat, dtype=tf.int32)
```
twuebi: `lengths` would be a more descriptive name

sticker-graph/sticker_graph/model.py
```diff
@@ -143,27 +159,23 @@ def subword_reprs(self):
         # Convert strings to a byte tensor.
         #
         # Shape: [batch_size, seq_len, subword_len]
-        subword_bytes = tf.strings.unicode_decode(
-            self.subwords, input_encoding='UTF-8')
-        subword_bytes_padded = subword_bytes.to_tensor(
-            0)[:, :, :self.args.subword_len]
+        with tf.device("/cpu:0"):
```
twuebi, line 162: is this necessary for `map_fn`?

sticker-graph/sticker_graph/model.py
```diff
@@ -143,27 +159,23 @@ def subword_reprs(self):
...
 
         # Get the lengths of the subwords. Only the last dimension should
         # be ragged, so no actual padding should happen.
         #
         # Shape: [batch_size, seq_len]
-        subword_lens = tf.math.minimum(
-            subword_bytes.row_lengths(
-                axis=-1).to_tensor(0),
-            self.args.subword_len)
+        with tf.device("/cpu:0"):
```
twuebi, line 167: see above

sticker-graph/sticker_graph/model.py
twuebi: unsure about the performance difference, but using something like `tf.count_nonzero` along the word axis could also return the word lengths without decoding and looping twice.

sticker-graph/sticker_graph/model.py
```diff
@@ -143,27 +159,23 @@ def subword_reprs(self):
...
         # Shape: [batch_size, seq_len, subword_len]
-        subword_bytes = tf.strings.unicode_decode(
-            self.subwords, input_encoding='UTF-8')
-        subword_bytes_padded = subword_bytes.to_tensor(
-            0)[:, :, :self.args.subword_len]
+        with tf.device("/cpu:0"):
+            subword_bytes = tf.cast(self.decode_words(self.subwords, self.args.subword_len), tf.int32)
 
         # Get the lengths of the subwords. Only the last dimension should
         # be ragged, so no actual padding should happen.
```
twuebi: Is the comment about the padding still up to date?

sticker-graph/sticker_graph/model.py
danieldk: I considered doing this, especially because the decoding is quite slow for some reason (didn't do any profiling). But I decided against it, since files are permitted to contain the NULL codepoint. It's bad for various reasons (e.g. because NULL is the terminator in C strings), but in the real-world files contain all kind of weird stuff. Of course, in such cases `count_nonzero` would give incorrect lengths.

One argument in favor of doing `count_nonzero` is that if tokens contain NULL all is probably lost anyway.

sticker-graph/sticker_graph/model.py
danieldk: On the GPU it's just too slow.

sticker-graph/sticker_graph/model.py
danieldk: The performance is the same, since the count_nonzero approach is shorter I have changed to that. I have thought a bit more about it, and I think that tokens will `NULL` characters are junk anyway, so it probably doesn't matter if we cut one or more characters by not counting `NULL`s.

sticker-graph/sticker_graph/model.py
danieldk: Removed.

twuebi:
Ci has some issue?

sticker-graph/sticker_graph/model.py
twuebi: I see, maybe put the subwords placeholder on the cpu too?

danieldk: Would be a useful thing to try. I'll make that a separate PR if that helps.

danieldk:
CI was having problems, restarted the job and it looks happy now. I guess that this PR is ready to go?

twuebi (approved):

PR #116, opened 2019-09-02
Add TF2 compatibility changes that are low-hanging fruit

danieldk:


sticker-graph/sticker_graph/model.py
```diff
@@ -24,21 +24,21 @@ def affine_transform(self, prefix, x, n_outputs):
...
         return tf.reshape(
-            tf.nn.xw_plus_b(
+            tf.compat.v1.nn.xw_plus_b(
                 x, w, b), [
                 batch_size, -1, n_outputs])
 
     def decode_word(self, word, max_subword_len):
-        bytes = tf.decode_raw(word, tf.uint8)[:max_subword_len]
-        return tf.pad(bytes, [[0, max_subword_len - tf.shape(bytes)[0]]])
+        bytes = tf.io.decode_raw(word, tf.uint8)[:max_subword_len]
```
sebpuetz, line 40: Unrelated, but isn't `bytes` a built-in?

sebpuetz (approved):
Didn't see anything odd, but maybe @twuebi is deeper into the matter.

sticker-graph/sticker_graph/model.py
danieldk: Not a keyword:
https://docs.python.org/3.7/reference/lexical_analysis.html#keywords

Which is probably why this is permitted. But it is indeed a built-in class, so this should probably be renamed.

sticker-graph/sticker_graph/write_helper.py
```diff
@@ -17,29 +17,33 @@ def create_graph(model, args):
...
+                tf.compat.v1.global_variables(), name='init')
 
-            tf.train.Saver(tf.global_variables())
+            tf.compat.v1.train.Saver(tf.compat.v1.global_variables())
 
             serialized_graph = session.graph_def.SerializeToString()
             serialized_graph_tensor = tf.convert_to_tensor(serialized_graph)
-            tf.contrib.summary.graph(serialized_graph_tensor, 0, name='graph_write')
-            tf.train.write_graph(
+            tf.contrib.summary.graph(
```
twuebi, line 44: `contrib` will be gone in 2.0, need to check if the method has moved elsewhere.

danieldk: `tf_upgrade_v2` gives suggestions for replacing `contrib.summary.graph`, but also mentions that it is not a 1:1 conversion. Maybe you could have a look at this sometime after this PR is merged? (I didn't want to touch the summary-writing parts of the graph, since you are the expert ;).)

twuebi: Should this PR be compatible with 2.0 already?

danieldk: No, it's just one step towards compatibility (and only the low-hanging fruit). E.g. for CRFs we need to introduce tensorflow-addons as a dependency. There are probably still other incompatibilities besides use of `contrib`.

I also didn't test yet whether the Rust `tensorflow` crate is compatible with TF2.

twuebi (approved):

danieldk:
Thanks for the reviews! I am doing a small test run on turing-sfb. If everything is fine, I'll merge.

danieldk:
Tested Dutch NER (since it does both RNNs + subword RNN), no regression in accuracy.

PR #120, opened 2019-09-23
Compute the unknown vector from the subquantizers for quantized embeddings

### LINKED ISSUE #119: Slow start with quantized embeddings
sticker uses the average of all embeddings for the unknown embedding. However, this has a bad side-effect for quantized embeddings. Since all the embeddings in the embedding matrix are reconstructed, loading models takes quite a while.

Possible workarounds:

* Precompute the unknown embedding and add it to the embedding matrix. I dislike this somewhat, since sticker would need specially prepared matrices again.
* Precompute the unknown embedding and store it somewhere outside the embedding matrix. Breaks existing models.
* Precompute the unknown embedding over a sample of the embeddings.
* Maybe it's enough to average all subquantizers, since they are the centroids of the embeddings? This would be very cheap. Problem: the original embeddings may not be distributed evenly over the clusters.

@twuebi @sebpuetz any opinions?

danieldk:
Fixes #119.

### DIFF
--- a/sticker/Cargo.toml
+++ b/sticker/Cargo.toml
@@ -14,7 +14,7 @@ license-file = "../LICENSE.md"
 [dependencies]
 conllx = "0.11"
 failure = "0.1"
-finalfusion = "0.9"
+finalfusion = "0.10.1"
 itertools = "0.8"
 ndarray = "0.12"
 ndarray-tensorflow = "0.2"
--- a/sticker/src/input.rs
+++ b/sticker/src/input.rs
@@ -5,7 +5,7 @@ use finalfusion::{
     chunks::vocab::VocabWrap,
     embeddings::Embeddings as FiFuEmbeddings,
 };
-use ndarray::Array1;
+use ndarray::{Array1, Axis};
 
 pub struct Embeddings {
     embeddings: FiFuEmbeddings<VocabWrap, StorageWrap>,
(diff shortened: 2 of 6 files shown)


sticker/Cargo.toml
```diff
@@ -14,7 +14,7 @@ license-file = "../LICENSE.md"
 [dependencies]
 conllx = "0.11"
 failure = "0.1"
-finalfusion = "0.9"
+finalfusion = "0.10.1"
```
sebpuetz, line 17: Is there motivation to fully specify the version here while leaving it at `0.10` for `sticker-utils`?

sticker/src/input.rs
```diff
@@ -26,11 +26,47 @@ impl Embeddings {
...
     fn from(embeddings: FiFuEmbeddings<VocabWrap, StorageWrap>) -> Self {
-        let mut unknown = Array1::zeros(embeddings.dims());
+        let mut unknown = match embeddings.storage() {
+            StorageWrap::QuantizedArray(_) | StorageWrap::MmapQuantizedArray(_) => {
+                let quantizer = match embeddings.storage() {
+                    StorageWrap::QuantizedArray(a) => a.quantizer(),
+                    StorageWrap::MmapQuantizedArray(a) => a.quantizer(),
+                    _ => unreachable!(),
+                };
+
```
sebpuetz, line 36: Couldn't this be written as a single `match` block returning `a.quantizer()`?

sebpuetz (approved):

sticker/Cargo.toml
danieldk: Only 0.10.1 has the `quantizer` getter, so we need that as a lower bound.

sticker/src/input.rs
danieldk: Good question, I don't think so, the problem is that `StorageWrap::QuantizedArray(a) | StorageWrap::MmapQuantizedArray(a)` would have incompatible types for `a`. And of course, in the outer block we first need to distinguish quantized embeddings from other embeddings.

sebpuetz: I meant something like:

```Rust
let quantizer = match embeddings.storage() {
    StorageWrap::QuantizedArray(a) => a.quantizer()
    StorageWrap::MmapQuantizedArray(a) => a.quantizer()
    _ => // deal with other embedding types
};

// deal with quantized embeddings
```

danieldk: Ah, I understand. But then the l2 normalization code could not be shared. And it would interleave quantizer and non-quantizer code.

danieldk:
Time for the next sticker release :). Better loading times, mmap'ing of quantized embeddings.

PR #124, opened 2019-10-07
Add the Pipeline type

danieldk:
This type is a wrapper around a `Vec` of `TaggerWrapper`s that offers a
`tag_sentences` method and applies the taggers sequentially.

### DIFF
--- a/sticker-utils/src/lib.rs
+++ b/sticker-utils/src/lib.rs
@@ -6,6 +6,9 @@ pub use crate::config::{
     Config, Embedding, EmbeddingAlloc, Embeddings, EncoderType, Input, Labeler, LabelerType,
 };
 
+mod pipeline;
+pub use crate::pipeline::Pipeline;
+
 mod progress;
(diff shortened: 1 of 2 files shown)


danieldk:
@DiveFish @twuebi  Anyone? This is a small, trivial change, the bigger change comes after this ;).

DiveFish:
How about adding a small test config and embeddings to make sure everything is loading correctly?

danieldk:
> How about adding a small test config and embeddings to make sure everything is loading correctly?

The problem is that we cannot include realistic embeddings and model in the repository. I want to go the same path as `sticker-python` and include extra tests that fetch external data. But that's more or less a separate project. I'll make an issue for that.

(And I am not sure if using small bogus embeddings and model is any good, it would only tell us that something is loaded that is a valid configuration, embeddings, and Tensorflow model. But it would not help catching any real regressions.)

danieldk:
Added as https://github.com/stickeritis/sticker/issues/125.

DiveFish (approved):

PR #126, opened 2019-10-09
Update sticker-tag and sticker-server to support pipelines

danieldk:
With this change, both programs accept multiple configuration files as
arguments. The models from the configuration files will be loaded and
used as a pipeline.

For `sticker-tag` this change means that the optional input/output
filename arguments have been replaced by the `--input` and `--output` options.

### DIFF
--- a/sticker-utils/src/app.rs
+++ b/sticker-utils/src/app.rs
@@ -1,6 +1,9 @@
 use clap::{App, AppSettings, Arg};
 
-static CONFIG: &str = "CONFIG";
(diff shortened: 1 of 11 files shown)


danieldk:
Added a small update to move the `readahead` and `batchsize` options to `sticker_pipeline_app`,

DiveFish (approved):
Looking ok to me!

danieldk:
@DiveFish Thanks for the review! Much appreciated!

PR #130, opened 2019-10-16
Add support for obtaining the top-k labels with their probabilities

danieldk:
- Adds a `TopK` trait and implements this for the Tensorflow `Tagger`.
- Factor out a `Tagger` method to get the top-k numerical labels,
  which is then both used for the `TopK` implementation and the `Tag`
  implementation.
- `CategorialEncoder::decode_without_inner` was added to just get the
  non-numerical labels. `CategoricalEncoder::decode` calls the inner
  encoder, which will update the sentence.
- Update `TaggerWrapper` to add a `top_k` method. Here, the labels are
  converted to strings. The goal of `TaggerWrapper` is to be devoid
  of types. Unfortunately, we do not have a trait for encodings
  (since they can be anything), so returning the individual labels
  as a `Box<Encoding>` is not possible. So, we return strings as
  an approximation.

---

I am not sure who the best reviewer for this code is. One of the reasons to add this is so that @DiveFish can inspect the top-k labels for interesting constructions. But @twuebi and @sebpuetz are more familiar with the code.

### DIFF
--- a/sticker-utils/src/tagger_wrapper.rs
+++ b/sticker-utils/src/tagger_wrapper.rs
@@ -10,8 +10,7 @@ use sticker::encoder::deprel::{RelativePOSEncoder, RelativePositionEncoder};
 use sticker::encoder::layer::LayerEncoder;
 use sticker::encoder::{CategoricalEncoder, SentenceDecoder};
 use sticker::tensorflow::{Tagger, TaggerGraph};
-use sticker::Tag;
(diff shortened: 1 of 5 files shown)


danieldk:
Fixed Travis CI clippy errors.

DiveFish (approved):
Very much looking forward to trying this out 👍

PR #131, opened 2019-10-16
Add a test using real-world models

### LINKED ISSUE #125: Add tests using real-world models
This could be done in a setup similar to what sticker-python uses.

danieldk:
This test uses German part-of-speech tagging and named entity
recognition models from our model repository. The output of the
pipeline using these models is then verified against the
current (almost-correct) output.

Run tests using: nix-build tests.nix

Fixes #125.


DiveFish (approved):
Thanks for adding this feature!

When run with ```NIXPKGS_ALLOW_UNFREE=1 nix-build tests.nix``` , it worked perfectly for me.

danieldk:
> Thanks for adding this feature!
> 
> When run with `NIXPKGS_ALLOW_UNFREE=1 nix-build tests.nix` , it worked perfectly for me.

Thanks for testing! And good point :), our models are marked as non-free. I don't really noticed, since I have my global nixpkgs set to permit non-free.

PR #140, opened 2019-10-19
Linear learning-rate warmup.

twuebi:
**Adds linear learning-rate warmup to sticker train.** 

The scheduler now needs to be passed into `run_epoch` since the warmup is _per-batch_ and not _per-epoch_. 

Not the scope of this PR: it may be useful to change `ExponentialDecay` to decay _per-batch_ instead of _per-epoch_.

### DIFF
--- a/sticker-utils/src/subcommands/train.rs
+++ b/sticker-utils/src/subcommands/train.rs
@@ -28,6 +28,7 @@ static LR_PATIENCE: &str = "LR_PATIENCE";
 static MAX_LEN: &str = "MAX_LEN";
 static CONTINUE: &str = "CONTINUE";
 static PATIENCE: &str = "PATIENCE";
+static WARMUP: &str = "WARMUP";
 static TRAIN_DATA: &str = "TRAIN_DATA";
 static VALIDATION_DATA: &str = "VALIDATION_DATA";
 static LOGDIR: &str = "LOGDIR";
@@ -36,6 +37,7 @@ pub struct LrSchedule {
     pub initial_lr: NotNan<f32>,
     pub lr_scale: NotNan<f32>,
     pub lr_patience: usize,
+    pub warmup_steps: usize,
 }
 
 pub struct TrainApp {
@@ -56,6 +58,7 @@ impl TrainApp {
             self.lr_schedule.initial_lr.into_inner(),
             self.lr_schedule.lr_scale.into_inner(),
             self.lr_schedule.lr_patience,
+            self.lr_schedule.warmup_steps,
         )
     }
 
@@ -69,8 +72,9 @@ impl TrainApp {
         trainer: &TaggerTrainer,
         file: &mut File,
         is_training: bool,
-        lr: f32,
-    ) -> (f32, f32)
+        lr_scheduler: &mut dyn LearningRateSchedule,
+        global_step: usize,
+    ) -> (f32, f32, usize)
     where
         E: SentenceEncoder,
         E::Encoding: Clone + Eq + Hash,
@@ -86,6 +90,7 @@ impl TrainApp {
 
         let mut dataset = ConllxDataSet::new(read_progress);
 
+        let mut global_step = global_step;
         let mut instances = 0;
         let mut acc = 0f32;
         let mut loss = 0f32;
@@ -94,6 +99,7 @@ impl TrainApp {
             .batches(encoder, vectorizer, config.model.batch_size, self.max_len)
             .or_exit("Cannot read batches", 1)
         {
+            let lr = lr_scheduler.compute_step_learning_rate(global_step);
             let tensors = batch.or_exit("Cannot read batch", 1).into_parts();
 
             let batch_perf = if is_training {
@@ -119,6 +125,7 @@ impl TrainApp {
             instances += n_tokens;
 
             if is_training {
+                global_step += 1;
                 save_scheduler.save(trainer, CompletedUnit::Batch).or_exit(
                     format!("Cannot save model for batch {}", save_scheduler.batch()),
                     1,
@@ -126,15 +133,15 @@ impl TrainApp {
             }
 
             progress_bar.set_message(&format!(
-                "batch loss: {:.4}, batch accuracy: {:.4}",
-                batch_perf.loss, batch_perf.accuracy
+                "lr: {:.1e} batch loss: {:.4}, batch accuracy: {:.4}",
+                lr, batch_perf.loss, batch_perf.accuracy
             ));
         }
 
         loss /= instances as f32;
         acc /= instances as f32;
 
-        (loss, acc)
+        (loss, acc, global_step)
     }
 
     fn train_model_with_encoder<E>(
@@ -170,14 +177,15 @@ impl TrainApp {
         let mut best_epoch = 0;
         let mut best_acc = 0.0;
         let mut last_acc = 0.0;
+        let mut global_step = 0;
 
         let mut lr_schedule = self.lr_schedule();
         let mut save_scheduler = self.save_schedule.to_save_scheduler("");
 
         for epoch in 0.. {
-            let lr = lr_schedule.learning_rate(epoch, last_acc);
+            let lr = lr_schedule.compute_epoch_learning_rate(epoch, last_acc);
 
-            let (loss, acc) = self.run_epoch(
+            let (loss, acc, global_step_after_epoch) = self.run_epoch(
                 config,
                 &mut save_scheduler,
                 &mut categorical_encoder,
@@ -185,9 +193,10 @@ impl TrainApp {
                 &trainer,
                 &mut train_file,
                 true,
-                lr,
+                &mut lr_schedule,
+                global_step,
             );
-
+            global_step = global_step_after_epoch;
             eprintln!(
                 "Epoch {} (train, lr: {:.4}): loss: {:.4}, acc: {:.4}",
                 epoch, lr, loss, acc
@@ -198,7 +207,7 @@ impl TrainApp {
                 1,
             );
 
-            let (loss, acc) = self.run_epoch(
+            let (loss, acc, _) = self.run_epoch(
                 config,
                 &mut save_scheduler,
                 &mut categorical_encoder,
@@ -206,7 +215,8 @@ impl TrainApp {
                 &trainer,
                 &mut validation_file,
                 false,
-                lr,
+                &mut lr_schedule,
+                global_step,
             );
 
             last_acc = acc;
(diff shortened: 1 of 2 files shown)


sebpuetz (commented):
I know this is how it was before, but I'm not a fan of `learning_rate` or now `compute_*_learning_rate` having side-effects such as incrementing step or decaying lr.

I think it's cleaner to have a dedicated `lr()` method, just returning the lr and additional `step()` and `epoch()` methods that change the scheduler's state.

sticker/src/tensorflow/lr.rs
```diff
@@ -66,23 +82,38 @@ impl ExponentialDecay {
...
+            warmup_steps,
         }
     }
 }
 
 impl LearningRateSchedule for ExponentialDecay {
-    fn learning_rate(&mut self, epoch: usize, _last_score: f32) -> f32 {
+    fn compute_batch_learning_rate(&mut self) -> f32 {
+        let lr = if self.step < self.warmup_steps {
+            self.initial_lr / (self.warmup_steps as f32) * self.step as f32
```
danieldk: Add parentheses to reduce ambiguity for the reader.

sticker/src/tensorflow/lr.rs
```diff
@@ -40,8 +45,11 @@ impl LearningRateSchedule for ConstantLearningRate {
 /// *lr = initial_lr * decay_rate ^ (epoch / decay_steps)*
 pub struct ExponentialDecay {
     initial_lr: f32,
+    lr: f32,
     decay_rate: f32,
     decay_steps: usize,
+    warmup_steps: usize,
+    step: usize,
```
danieldk: `step` and `warmup_steps` get confusing here, because `decay_steps` refers to the number of epochs, while `step` and `warmup_steps` refer to batches?

sticker/src/tensorflow/lr.rs
```diff
@@ -66,23 +82,38 @@ impl ExponentialDecay {
...
+            self.initial_lr / (self.warmup_steps as f32) * self.step as f32
+        } else {
+            self.lr
+        };
+        self.step += 1;
+        lr
+    }
+
+    fn compute_epoch_learning_rate(&mut self, epoch: usize, _last_score: f32) -> f32 {
+        self.step += 1;
```
danieldk: Should `step` be increased here as well, since it is already increased for batches?

danieldk (changes requested):

danieldk:
> I know this is how it was before, but I'm not a fan of `learning_rate` or now `compute_*_learning_rate` having side-effects such as incrementing step or decaying lr.

My initial idea (and I think implementation) was to have stateless/immutable schedules. But this sort of broke down with the plateau scheduler, since you need to keep track of your lr, since it is not a direct function of the number of epochs.

> I think it's cleaner to have a dedicated `lr()` method, just returning the lr and additional `step()` and `epoch()` methods that change the scheduler's state.

Well, I think we should still strive for immutable schedules as much as possible. So I think here it means that the caller has to give the overall number of batches that were processed as an argument, rather than keeping track of the step internally.

sticker/src/tensorflow/lr.rs
twuebi: I think decay steps should actually refer to batches too (e.g. https://www.tensorflow.org/api_docs/python/tf/compat/v1/train/exponential_decay). In TF terminology a step is usually a single call to the train op.

sticker/src/tensorflow/lr.rs
twuebi: Thanks, that shouldn't be there.

sticker/src/tensorflow/lr.rs
danieldk: A lot of this code has a heritage from before Tensorflow (through the old Go versions of dparnn/toponn). I agree that using batches would be better, since it would decay in a more fine-grained manner. However, it would also mean that the meaning of the parameters related to this schedule would change.

At any rate, I think words like *iteration* or *step* are very confusing, especially since we have the non-ambiguous terms *batch* and *epoch*. So I propose to rename `warmup_steps` and `step` to `warmup_batches` and `batch`. We can then later change the behavior of decay to be in number of batches as well.

twuebi: When I read batch, I'd expect it to be about the n-th batch in the epoch and not about the total number of processed batches.

In tensorflow and pytorch step describes a single iteration in the optimization process:

* the `global_step` argument of `tf.train.AdamOptimizer.minimize` (https://www.tensorflow.org/api_docs/python/tf/compat/v1/train/AdamOptimizer#minimize) or `tf.train.exponential_decay` (https://www.tensorflow.org/api_docs/python/tf/compat/v1/train/exponential_decay)
* in pytorch applying the gradients happens via calling `step` on the optimizer (https://pytorch.org/docs/stable/optim.html)

Would `global_step` or `global_train_step` be an acceptable name?

danieldk: `global_step` then. But maybe you could then rename `decay_steps` to `decay_epochs` as part of this PR to avoid confusion?

(The behavior of decay should be changed in a separate PR.)

twuebi: Not happy about `decay_epochs` either. If `staircase == true` it describes the width of a "step" in the decay stairs but more importantly it also influences the slope. Both properties don't really seem to be reflected in `decay_steps` or `decay_epochs` which I would either read as _decay-steps-before-zero-lr_ or _decay-steps-before-constant-lr_.

Unfortunately I cannot come up with a better name either. For now, I'll use `decay_epochs` to make clear it's about epochs, not steps. Maybe we'll come up with something better when changing the decay-behavior.

twuebi:
> Well, I think we should still strive for immutable schedules as much as possible. So I think here it means that the caller has to give the overall number of batches that were processed as an argument, rather than keeping track of the step internally.

Should the global step correspond to the one used in the graph or be per-run?

danieldk:
> Should the global step correspond to the one used in the graph or be per-run?

Could you explain what you mean by 'in the graph'? Do we store them?

So, the implication would be that we could continue training with the same LR?

twuebi:
> Could you explain what you mean by 'in the graph'? Do we store them?

We keep a global step variable in the graph for summary writing. It's created via `tf.train.create_or_get_global_step()` and passed to `optimizer.apply_gradients` / `optimizer.minimize` which increments it.


> So, the implication would be that we could continue training with the same LR?

Yes, which may be desirable in some cases (continue training on the same data) and not in others (pre-train -> fine-tune).

danieldk:
> Yes, which may be desirable in some cases (continue training on the same data) and not in others (pre-train -> fine-tune).

Let's not do it (so use the global step of the current run). Implicit state is always hard to reason about and easy to overlook.

twuebi:
> > Yes, which may be desirable in some cases (continue training on the same data) and not in others (pre-train -> fine-tune).
> 
> Let's not do it (so use the global step of the current run). Implicit state is always hard to reason about and easy to overlook.

Sure.

Something else to think about (not directly related to this PR) : we're currently keeping Adam's state across restarts. This has two effects a) it increases checkpoint size by `2*num_model_parameters` b) we're reusing it for fine-tuning.

sticker/src/tensorflow/lr.rs
```diff
@@ -8,7 +8,9 @@ use std::f32;
 /// at a given epoch.
 pub trait LearningRateSchedule {
     /// Compute the learning rate for an epoch.
-    fn learning_rate(&mut self, epoch: usize, last_score: f32) -> f32;
+    fn compute_epoch_learning_rate(&mut self, epoch: usize, last_score: f32) -> f32;
+    /// Compute the learning rate for the current batch.
+    fn compute_step_learning_rate(&mut self, global_step: usize) -> f32;
```
twuebi, line 13: Changed this from `compute_batch_learning_rate` to `compute_step_learning_rate` to be consistent with `global_step` and `warmup_steps`.

sticker/src/tensorflow/lr.rs
```diff
@@ -51,7 +58,15 @@ impl ExponentialDecay {
     /// If `staircase` is true, the exponent of the decay is
     /// computed using integer division. This has the effect that
     /// the learning rate only changes every `decay_steps` steps.
-    pub fn new(initial_lr: f32, decay_rate: f32, decay_steps: usize, staircase: bool) -> Self {
+    /// If `warmup_steps` > 0, the learning rate is linearly scaled
+    /// for `warmup_steps` from 0 -> `initial_lr`.
+    pub fn new(
+        initial_lr: f32,
+        decay_rate: f32,
+        decay_steps: usize,
```
danieldk: `decay_epochs`

sticker/src/tensorflow/lr.rs
```diff
@@ -66,23 +81,34 @@ impl ExponentialDecay {
...
+            (self.initial_lr / (self.warmup_steps as f32)) * global_step as f32
+        } else {
+            self.lr
+        }
+    }
+
+    fn compute_epoch_learning_rate(&mut self, epoch: usize, _last_score: f32) -> f32 {
         let exponent = if self.staircase {
-            (epoch / self.decay_steps) as f32
+            (epoch / self.decay_epochs) as f32
```
danieldk, line 105: For after this PR: once decay is also in batches/steps, we should probably start computing the exponent as `(steps - warmup_steps) / decay_steps`, so that decay only starts after the warmup.

danieldk (approved):
One small arg name change request, good to go after that.

sticker-utils/src/subcommands/train.rs
```diff
@@ -170,22 +176,24 @@ impl TrainApp {
...
 
         let mut lr_schedule = self.lr_schedule();
         let mut save_scheduler = self.save_schedule.to_save_scheduler("");
 
         for epoch in 0.. {
-            let lr = lr_schedule.learning_rate(epoch, last_acc);
+            let lr = lr_schedule.compute_epoch_learning_rate(epoch, last_acc);
 
-            let (loss, acc) = self.run_epoch(
+            let (loss, acc, global_step) = self.run_epoch(
```
twuebi: This global_step gets dropped at the end of the scope, leading to a learning rate that resets after every epoch.

danieldk: Whoops! Also explains why the global_step above didn't need to be `mut`. Easy to fix though ;).

danieldk: Maybe we should also start showing the current LR in the progress, besides the loss? Would make it easier to see what is going on.

twuebi: Tensorboard :). But I agree, it's also nice to see it without resorting to a browser.

twuebi:
Added the lr to the progress, I opted for scientific notation since there are usually lots of zeros. This good to go?

danieldk (approved):

PR #142, opened 2019-10-20
Sentence iterator adapter.

twuebi:
This PR introduces a trait `SentenceIter` that provides adapters for `conllx::io::Sentences`. 

This trait makes it easier to introduce other operations such as shuffling via a buffer or local sorting.

***

In initial experiments buffered shuffling gave better results while converging faster than linear sampling.

***

The design is modelled after https://github.com/finalfusion/finalfrontier/blob/master/src/deps.rs

### DIFF
--- a/sticker/src/tensorflow/dataset.rs
+++ b/sticker/src/tensorflow/dataset.rs
@@ -2,8 +2,9 @@ use std::hash::Hash;
 use std::io::{BufReader, Read, Seek, SeekFrom};
 use std::usize;
 
-use conllx::io::{ReadSentence, Reader, Sentences};
-use failure::Fallible;
+use conllx::graph::Sentence;
+use conllx::io::{ReadSentence, Reader};
+use failure::{Error, Fallible};
 
 use super::tensor::{LabelTensor, TensorBuilder};
 use crate::encoder::{CategoricalEncoder, SentenceEncoder};
@@ -45,6 +46,27 @@ impl<R> ConllxDataSet<R> {
     pub fn new(read: R) -> Self {
         ConllxDataSet(read)
     }
+
+    /// Returns an `Iterator` over `Result<Sentence, Error>`.
+    ///
+    /// Depending on the parameters the returned iterator filters
+    /// sentences by their lengths or returns the sentences in
+    /// sequence without filtering them.
+    ///
+    /// If `max_len` == `usize::MAX`, no filtering is performed.
+    fn get_sentence_iter<'a>(
+        reader: R,
+        max_len: usize,
+    ) -> Box<dyn Iterator<Item = Result<Sentence, Error>> + 'a>
+    where
+        R: ReadSentence + 'a,
+    {
+        if max_len < usize::MAX {
+            Box::new(reader.sentences().filter_by_len(max_len))
+        } else {
+            Box::new(reader.sentences())
+        }
+    }
 }
 
 impl<'a, 'ds, E, R> DataSet<'a, E> for &'ds mut ConllxDataSet<R>
@@ -53,7 +75,7 @@ where
     E::Encoding: 'a + Clone + Eq + Hash,
     R: Read + Seek,
 {
-    type Iter = ConllxIter<'a, E, Reader<BufReader<&'ds mut R>>>;
+    type Iter = ConllxIter<'a, E, Box<dyn Iterator<Item = Result<Sentence, Error>> + 'ds>>;
 
     fn batches(
         self,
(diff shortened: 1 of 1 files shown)


sticker/src/tensorflow/dataset.rs
```diff
@@ -70,31 +71,29 @@ where
...
     batch_size: usize,
-    max_len: usize,
     encoder: &'a mut CategoricalEncoder<E, E::Encoding>,
     vectorizer: &'a SentVectorizer,
-    sentences: Sentences<R>,
+    sentences: I,
 }
 
-impl<'a, E, R> Iterator for ConllxIter<'a, E, R>
+impl<'a, E, I> Iterator for ConllxIter<'a, E, I>
```
sebpuetz, line 113: This actually is a `ConllxBatcher` or `ConllxBatchIter`, no?

sebpuetz (dismissed):
I think the boxing-indirection is necessary and I'm pretty sure it doesn't bring a considerable performance penalty.

Looks good to me.

sticker/src/tensorflow/dataset.rs
```diff
@@ -151,3 +146,65 @@ where
...
+
+impl<I> Iterator for LengthFilter<I>
+where
+    I: Iterator<Item = Result<Sentence, Error>>,
+{
+    type Item = Result<Sentence, Error>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while let Some(sent) = self.inner.next() {
+            let len = sent.as_ref().map(|s| s.len()).unwrap_or(0);
```
danieldk, line 204: `sent` is a `Result<Sentence, Error>`, this will result in `len == 0` when the `Result` is `Err`. This is fine here, since the iterator will properly return `Some(Err(_))` below. But it is probably good to add a small comment above this line pointing that out.

sticker/src/tensorflow/dataset.rs
```diff
@@ -151,3 +146,65 @@ where
...
+    }
+}
+
+/// Returns an `Iterator` over `Result<Sentence, Error>`.
+///
+/// Depending on the parameters the returned iterator filters
+/// sentences by their lengths or returns the sentences in
+/// sequence without filtering them.
+///
+/// If `max_len` == `usize::MAX`, no filtering is performed.
```
danieldk: I know that I am the guilty party here, but interface-wise, it would be nicer to take an `Option`, so that `None` can be used for no filtering. This can be a separate PR.

sticker/src/tensorflow/dataset.rs
```diff
@@ -151,3 +146,65 @@ where
...
+fn get_sentence_iter<'a, R>(
```
danieldk: This should be an association function/static method of `ConllxDataSet`.

danieldk (changes requested):
Two small requested changes.

sticker/src/tensorflow/dataset.rs
twuebi: Added a comment

sticker/src/tensorflow/dataset.rs
```diff
@@ -45,6 +46,27 @@ impl<R> ConllxDataSet<R> {
...
+    /// sentences by their lengths or returns the sentences in
+    /// sequence without filtering them.
+    ///
+    /// If `max_len` == `usize::MAX`, no filtering is performed.
+    fn get_sentence_iter<'a>(
+        reader: R,
+        max_len: usize,
+    ) -> Box<dyn Iterator<Item = Result<Sentence, Error>> + 'a>
+    where
+        R: ReadSentence + 'a,
```
twuebi, line 62: There are two options here:

Option a) is the current state, we don't have a local parameter on `get_sentence_iter` but use `R` from `ConllxDataSet`.

Option b) introduces a local parameter `S` on `get_sentence_iter` and needs to specify the type of the `R` on `ConllxDataSet` when calling `get_sentence_iter`:

```rust
    fn get_sentence_iter<'a,S>(
        reader: S,
        max_len: usize,
    ) -> Box<dyn Iterator<Item = Result<Sentence, Error>> + 'a>
    where
        S: ReadSentence + 'a,
    {...}

  
    fn batches(...) -> Fallible<Self::Iter> {

        Ok(ConllxIter {
            batch_size,
            encoder,
            sentences: ConllxDataSet::<R>::get_sentence_iter(reader, max_len),
            vectorizer,
        })
    }
}
```

danieldk: I think the current approach is good enough. We could always relax the type parameters when necessary.

danieldk (approved):

PR #145, opened 2019-10-22
Make exponential learning-rate decay per step.

twuebi:
Make the decay in the `ExponentialRate` schedule based on the global step instead of the training epoch.

### DIFF
--- a/sticker/src/tensorflow/lr.rs
+++ b/sticker/src/tensorflow/lr.rs
@@ -42,12 +42,12 @@ impl LearningRateSchedule for ConstantLearningRate {
 /// exponentionally over time. To be specific, the learning rate is
 /// calculated as follows:
 ///
-/// *lr = initial_lr * decay_rate ^ (epoch / decay_steps)*
+/// *lr = initial_lr * decay_rate ^ (global_step / decay_steps)*
 pub struct ExponentialDecay {
     initial_lr: f32,
     lr: f32,
     decay_rate: f32,
-    decay_epochs: usize,
+    decay_steps: usize,
     warmup_steps: usize,
     staircase: bool,
 }
@@ -57,13 +57,13 @@ impl ExponentialDecay {
     ///
     /// If `staircase` is true, the exponent of the decay is
     /// computed using integer division. This has the effect that
-    /// the learning rate only changes every `decay_epochs` epochs.
+    /// the learning rate only changes every `decay_steps` steps.
     /// If `warmup_steps` > 0, the learning rate is linearly scaled
     /// for `warmup_steps` from 0 -> `initial_lr`.
     pub fn new(
         initial_lr: f32,
         decay_rate: f32,
-        decay_epochs: usize,
+        decay_steps: usize,
         staircase: bool,
         warmup_steps: usize,
     ) -> Self {
@@ -76,7 +76,7 @@ impl ExponentialDecay {
             "The decay rate must be in (0, 1)."
         );
         assert!(
-            decay_epochs > 0,
+            decay_steps > 0,
             "The number decay steps should be non-zero."
         );
 
@@ -84,7 +84,7 @@ impl ExponentialDecay {
             lr: initial_lr,
             initial_lr,
             decay_rate,
-            decay_epochs,
+            decay_steps,
             staircase,
             warmup_steps,
         }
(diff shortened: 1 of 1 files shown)


sticker/src/tensorflow/lr.rs
```diff
@@ -94,22 +94,23 @@ impl ExponentialDecay {
...
-            (self.initial_lr / (self.warmup_steps as f32)) * global_step as f32
-        } else {
-            self.lr
+            return (self.initial_lr / (self.warmup_steps as f32)) * global_step as f32;
         }
-    }
 
-    fn compute_epoch_learning_rate(&mut self, epoch: usize, _last_score: f32) -> f32 {
+        // start decay after warmup
+        let decay_epoch = global_step - self.warmup_steps;
```
sebpuetz: `decay_step`?

sticker/src/tensorflow/lr.rs
```diff
@@ -57,13 +57,13 @@ impl ExponentialDecay {
     ///
     /// If `staircase` is true, the exponent of the decay is
     /// computed using integer division. This has the effect that
-    /// the learning rate only changes every `decay_epochs` epochs.
+    /// the learning rate only changes every `decay_steps` steps.
```
sebpuetz, line 60: Not part of this change, but would generally be informative: Note that `decay_steps` not only influences the width of the steps but also directly affects the decay rate.

sticker/src/tensorflow/lr.rs
twuebi: thanks

sticker/src/tensorflow/lr.rs
twuebi: Agreed. I'd also prefer a more descriptive name than `decay_steps` which mostly makes sense in the `staircase`-case. FWIW the formula is included in the doc of the struct:

> *lr = initial_lr * decay_rate ^ (global_step / decay_steps)*

twuebi: Anyways, the names we got now mirror those of https://www.tensorflow.org/api_docs/python/tf/compat/v1/train/exponential_decay. 

OTOH, PyTorch-exponential decay completely avoids `decay_steps` by not offering stair-casing:

```py
class ExponentialLR(_LRScheduler):
    """Set the learning rate of each parameter group to the initial lr decayed
    by gamma every epoch. When last_epoch=-1, sets initial lr as lr.

    Args:
        optimizer (Optimizer): Wrapped optimizer.
        gamma (float): Multiplicative factor of learning rate decay.
        last_epoch (int): The index of last epoch. Default: -1.
    """

    def __init__(self, optimizer, gamma, last_epoch=-1):
        self.gamma = gamma
        super(ExponentialLR, self).__init__(optimizer, last_epoch)

    def get_lr(self):
        return [base_lr * self.gamma ** self.last_epoch
                for base_lr in self.base_lrs]
```

I am now wondering whether we should just opt for the simple version.

danieldk: If simple means no staircasing, no please leave staircasing in. I had some models in the past where staircasing improved performance a bit.

twuebi: With _per-epoch_ or _per-batch_ decay? 
I guess it doesn't hurt to have it.

sticker/src/tensorflow/lr.rs
```diff
@@ -94,22 +94,23 @@ impl ExponentialDecay {
...
-            (self.initial_lr / (self.warmup_steps as f32)) * global_step as f32
-        } else {
-            self.lr
+            return (self.initial_lr / (self.warmup_steps as f32)) * global_step as f32;
         }
-    }
 
-    fn compute_epoch_learning_rate(&mut self, epoch: usize, _last_score: f32) -> f32 {
+        // start decay after warmup
+        let decay_steps = global_step - self.warmup_steps;
```
danieldk: Isn't there an off-by-one here? If `global_step == warmup_steps`, then there is a division by zero below?

danieldk (changes requested):

sticker/src/tensorflow/lr.rs
twuebi: No, this is dividing 0 by `decay_steps`.

twuebi: Ah I see where the confusion is coming from. We have `self.decay_steps` and this local variable which have almost identical names. I'll change that.

sticker/src/tensorflow/lr.rs
danieldk: That was per epoch. But now it can be per batch. But in the end it shouldn't matter to much. One has to adjust the hyper parameters anyway to get the same schedule as before (old steps * number of batches in the training data).

twuebi: Maybe I am missing something, but right now exponential decay is only available when using sticker as a library?

sticker/src/tensorflow/lr.rs
danieldk: Ah, yes! It's much clearer now...

danieldk (approved):

sticker/src/tensorflow/lr.rs
danieldk: Yes. Plateau replaced it as the default, but I never added an option to switch to exponential decay.

PR #148, opened 2019-10-23
Split savers in separate data structures, add BestEpochSaver

### LINKED ISSUE #65: Option to only save with improvements
Currently, we keep all saved model files. With ~100 epochs per model, this produces a pretty big amount of files where the size depends on model type / size.

It would be nice to have an option to only save the new parameters if `cur_performance > best_performance` or to only keep the last `n_patience` checkpoints.

danieldk:
There used to be just one saver, `SaveScheduler`, with a `SaveSchedule`
enum to decide the saving schedule. The problem with this approach is
that the information that `SaveScheduler` had to keep track of increased
with each new `SaveSchedule`.

This change defines a `Save` trait and splits the different schedules
into separate saver data types that each implement the `Save` trait.
Additionally, the `BestEpochSaver` schedule is added that only saves the
model for an epoch when the model has improved over all previous
epochs.

Fixes #65.

---

For now, I have left the pretrain schedule the same (though I still have to test it). I will later add a schedule that can be added to pretrain for only saving improved (on the validation data) batches.

### DIFF
--- a/sticker-utils/src/save.rs
+++ b/sticker-utils/src/save.rs
@@ -1,86 +1,166 @@
-use failure::Fallible;
+use std::marker::PhantomData;
 
+use failure::{Fallible, ResultExt};
 use sticker::tensorflow::TaggerTrainer;
 
 #[derive(Copy, Clone, Eq, PartialEq)]
-pub enum CompletedUnit {
-    /// A batch is completed.
-    Batch,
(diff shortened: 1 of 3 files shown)


danieldk:
Fixed clippy unhappiness. Tested with pretraining, works as expected.

twuebi (approved):

PR #152, opened 2019-10-23
Also add the --batchsize option to training apps

### LINKED ISSUE #151: Batch size in config
Changing the batch size on the fly is a nice convenience when experimenting. Does something speak against moving it out of the config file?

danieldk:
- The pipeline apps were already using the --batchsize option in place
  of the sticker configuration file. Make things consistent again by
  doing the same in the train apps.
- Add a warning that the 'model.batch_size' option is not used anymore
  if it is present in the configuration.

Fixes #151.

### DIFF
--- a/sticker/src/tensorflow/tagger.rs
+++ b/sticker/src/tensorflow/tagger.rs
@@ -23,9 +23,8 @@ use crate::{SentVectorizer, Tag, TopK, TopKLabels};
 #[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
 #[serde(deny_unknown_fields)]
 pub struct ModelConfig {
-    /// Model batch size, should be kept constant between training and
-    /// prediction.
-    pub batch_size: usize,
+    /// Model batch size (unused).
+    pub batch_size: Option<usize>,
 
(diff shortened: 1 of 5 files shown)


danieldk:
Still have to test this ;).

sticker/src/tensorflow/tagger.rs
```diff
@@ -25,7 +25,7 @@ use crate::{SentVectorizer, Tag, TopK, TopKLabels};
 pub struct ModelConfig {
     /// Model batch size, should be kept constant between training and
```
twuebi: Why should batch size be constant between train and prediction?

danieldk: I have no idea. Maybe this was read from the TOML configuration and then hardcoded in the graph at some point? I think initially the major dimension was a concrete batch size and not `None`. Then if you'd change the value, everything would crash and burn in prediction. IIRC old Tensorflow versions do not have dynamic shapes and this has gone through Go + Caffe -> Go + Tensorflow -> Rust + Tensorflow -> Rust + Tensorflow Rust training and probably through every public Tensorflow version ever ;).

twuebi (approved):

PR #158, opened 2019-10-30
 Change SentenceEncoder::encode to take &self immutably

### LINKED ISSUE #155: Make SentenceEncoder::encode take self immutably
This is currently making the implementation of alternating train/validation steps in `pretrain` difficult, since we cannot have two mutable references to the categorical encoder. The `&mut self` is currently required because the categorical encoder updates a `Numberer`.

We could switch to interior mutability for `CategoricalEncoder`, however this makes this encoder unsharable between threads (and thus not work in sticker server) unless it's wrapped in an `Arc`.

Possible solutions:

* Wrap the numberer in an Arc (what is the performance impact)?
* Make mutable and immutable flavors of this encoder.

danieldk:
This requires changes accross the board, since CategoricalEncoder used
a numberer immutably. This change adds a type parameter to the
CategoricalEncoder trait that should implement the Mutability trait.
There are two crate-private data structures that implement this trait:

- Immutable: wraps Numberer directly, never updates the Numberer. If a
  value is not in the Numberer, it returns `None`.
- Mutable: wraps Numberer in a RefCell. Updates the Numberer when new
  values are encountered.

Two corresponding type aliases are introduced:
ImmutableCategoricalEncoder and MutableCategoricalEncoder. The
ImmutableCategoricalEncoder returns the special value `0` when an
encoding is unknown.

Throughout sticker we now only use ImmutableCategoricalEncoder, so it
should *always* return a known number, thereby also fixing #28.

Fixes #155

---

Currently a draft PR, since this still needs testing.


sticker/src/encoder/categorical/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+/// An immutable categorical encoder
+///
+/// This encoder does not add new encodings to the encoder. If the
+/// number of an unknown encoding is looked up, the special value `0`
+/// is used.
+pub type ImmutableCategoricalEncoder<E, V> = CategoricalEncoder<E, V, mutability::Immutable<V>>;
+
+/// A mutable categorical encoder
+///
+/// This encoder does adds new encodings to the encoder when
```
twuebi: `s/does//`

danieldk:
Doing a test run now, at least there are validation losses :).

sticker/src/encoder/categorical/mod.rs
danieldk: Thanks! Fixed.

danieldk:
> Doing a test run now, at least there are validation losses :).

Works as it did before, so I converted this into a proper non-review PR. I will also test it with an existing model, just to be 100% certain that nothing breaks.

sticker/src/encoder/categorical/mod.rs
```diff
@@ -1,44 +1,122 @@
...
+        }
+
+        fn number(&self, value: V) -> Option<usize> {
+            Some(self.0.borrow_mut().add(value))
+        }
+
+        fn value(&self, number: usize) -> Option<V> {
+            self.0.borrow().value(number).cloned()
+        }
+    }
```
twuebi, line 67: Maybe `MutableNumberer` and `ImmutableNumberer` would be more descriptive names than `Mutable` and `Immutable`.  Perhaps also rename `Mutability` to something like `Number` since this seems to be the primary function of the trait.

Then we would have a trait `Number` implemented on `MutableNumberer` and `ImmutableNumberer`

danieldk: Sounds good.

twuebi (approved):

PR #159, opened 2019-10-30
Add shuffling adapter to SentenceIter

twuebi:
This adds a buffered shuffling adapter to SentenceIter. The adapter first fills a buffer with items. After the buffer is filled, incoming items are swapped with a random element from the buffer. The random element is the next item. The buffer size controls the locality of the shuffling. If buffer_size is larger than the number of sentences in the dataset this is equivalent to uniform sampling over the full data set.


sticker/src/tensorflow/dataset.rs
```diff
@@ -209,3 +231,61 @@ where
...
+                sent
+            }
+            None => {
+                // if the buffer is empty, we know that `inner` is exhausted and return None.
+                let mut sent = self.buffer.pop()?;
+
+                // if the buffer is empty after taking the last item, skip swapping the item
+                // to avoid indexing into an empty Vec.
+                if !self.buffer.is_empty() {
+                    let buffer_len = self.buffer.len();
```
danieldk: Use `Vec::swap_remove`.

sticker/src/tensorflow/dataset.rs
```diff
@@ -209,3 +231,61 @@ where
...
+                if self.buffer.len() == self.buffer_size {
+                    break;
+                }
+            }
+        }
+
+        let sent = match self.inner.next() {
+            Some(mut sent) => {
+                let buffer_len = self.buffer.len();
+                std::mem::swap(
```
danieldk: This seems to be biased in two ways:

1. It should (probably?) be possible to randomly draw the new element (`sent` here).
2. Using the remainder operator to index into a range introduces bias. E.g., consider a buffer of length 2 and a (hypothetical) random number generator that produces numbers `[0,2]`. Index `0` has twice the probability of being drawn as index `1`. Use [`Uniform`](https://docs.rs/rand/0.7.2/rand/distributions/uniform) or [`Rng::gen_range`](https://docs.rs/rand/0.7.2/rand/trait.Rng.html#method.gen_range).

sticker/src/tensorflow/dataset.rs
```diff
@@ -209,3 +231,61 @@ where
...
+/// An Iterator adapter performing local shuffling.
+///
+/// Fills a buffer with size `buffer_size` on the first call. Subsequent
+/// calls swap the next incoming item with a random element from the
+/// buffer and return the random element.
+pub struct Shuffled<I> {
+    inner: I,
+    buffer: Vec<Result<Sentence, Error>>,
+    buffer_size: usize,
+    shuffler: XorShiftRng,
```
danieldk: Rename field `shuffler` to `rng`.

sticker/src/tensorflow/dataset.rs
```diff
@@ -209,3 +231,61 @@ where
...
 }
+
+/// An Iterator adapter performing local shuffling.
+///
+/// Fills a buffer with size `buffer_size` on the first call. Subsequent
+/// calls swap the next incoming item with a random element from the
+/// buffer and return the random element.
+pub struct Shuffled<I> {
+    inner: I,
+    buffer: Vec<Result<Sentence, Error>>,
```
danieldk: Maybe this should be `Vec<Sentence>` so that I/O errors fail as early as possible?

danieldk (changes requested):
Still have to do a detailed review, but here are some initial lower-hanging fruit comments.

sticker/src/tensorflow/dataset.rs
twuebi: The `Item` type of `SentenceIter` is `Result<Sentence, Error>>`. If we want to fail as early as possible, we'd need to change that type.

sticker/src/tensorflow/dataset.rs
twuebi: Not sure which one to use here. We are sampling quite often and the range stays the same until `inner` is exhausted, which probably means `Uniform` would be the best choice. OTOH, we'd need to take care of the possible out-of-bounds that can occur if we continue to sample from `Uniform` after `inner` is exhausted.

danieldk: `Rng::gen_range` should be ok. It calls this implementation:

https://docs.rs/rand/0.7.2/src/rand/distributions/uniform.rs.html#420

The initialization part is really cheap:  a subtraction to get a range, then: count leading zeros, bit shift, subtraction (for which x86_64 has instructions) to determine the largest multiple of the range.

sticker/src/tensorflow/dataset.rs
danieldk: The iterator needs that type, due to laziness. But the buffering is a greedy, so it could unwrap sentences and if it hits an `Err` return `Some(Err(...))` immediately (thus rewrapping the error).

sticker/src/tensorflow/dataset.rs
```diff
@@ -209,3 +231,51 @@ where
...
+}
+
+impl<I> Iterator for Shuffled<I>
+where
+    I: Iterator<Item = Result<Sentence, Error>>,
+{
+    type Item = Result<Sentence, Error>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.buffer.is_empty() {
```
danieldk, line 254: `if self.buffer.len() < self.buffer_size` is probably more correct here (especially when failing eagerly as suggested in the other comment).

twuebi: That'd mean we go `self.buffer_size  - 1` times into this if-branch to check whether `self.inner` has something to offer.

sticker/src/tensorflow/dataset.rs
twuebi: Not sure if I'm understanding this. 

If `self.buffer` is a `Vec<Sentence>`, we need to unwrap the incoming sentences, check for Err, return if there is one and then re-wrap the non-Err values in `Ok` to comply with the type of `SentenceIter`. 

```rust
match sent {
    Ok(sent) => buffer.push(sent),
    e => return Some(e)
}

sent = get_random_sent();

Some(Ok(sent))
```

Or did you mean something like:

```rust
            while let Some(sent) = self.inner.next() {
                if sent.is_err() {
                    return sent
                }
                self.buffer.push(sent);
                if self.buffer.len() == self.buffer_size {
                    break;
                }
            }
```
In both cases, we eliminated the Err variant of the type of `SentenceIter`.

sticker/src/tensorflow/dataset.rs
danieldk: Why would you go that number of times into the `if` branch? Only the first time, right, when the buffer is filled?

twuebi: If `self.inner` is exhausted, `self.buffer.len()`will be smaller than `self.buffer_size` for the remaining calls to `self.next`.

sticker/src/tensorflow/dataset.rs
danieldk: The former, since the latter would still require `buffer` to be `Vec<Result<...>>`. Rewrapping in `Ok` should be very cheap. Since there is a swap (which is a copy) anyway, it will probably be swapped in-place on the stack after optimization.

sticker/src/tensorflow/dataset.rs
danieldk: Ah, sorry, I misread it as going into the branch for the the overall number of items - 1! I was thinking about the situation that somebody hits an `Err` but decides to ask for the next item anyway. But I guess people shouldn't do that ;).

twuebi: So keep `self.buffer.is_empty`?

danieldk (approved):

PR #160, opened 2019-10-31
Shuffled: factor out Vec with random draws

danieldk:
- New items are also included in random draws.
- Use a unit test to verifify proper behavior of random draws.

### DIFF
--- a/sticker/src/tensorflow/dataset.rs
+++ b/sticker/src/tensorflow/dataset.rs
@@ -5,10 +5,11 @@ use std::usize;
 use conllx::graph::Sentence;
 use conllx::io::{ReadSentence, Reader};
 use failure::{Error, Fallible};
-use rand::{Rng, SeedableRng};
+use rand::SeedableRng;
 use rand_xorshift::XorShiftRng;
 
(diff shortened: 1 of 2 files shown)


twuebi (dismissed):
Looks good.

This already included the new item, right?

```rust
let sent = match self.inner.next() {
            Some(sent) => {
                match sent {
                    Ok(sent) => self.buffer.push(sent),
                    Err(err) => return Some(Err(err)),
                }
                let buffer_len = self.buffer.len();
                self.buffer.swap_remove(self.rng.gen_range(0, buffer_len))
            }
```

danieldk:
> Looks good.
> 
> This already included the new item, right?
> 
> ```rust
> let sent = match self.inner.next() {
>             Some(sent) => {
>                 match sent {
>                     Ok(sent) => self.buffer.push(sent),
>                     Err(err) => return Some(Err(err)),
>                 }
>                 let buffer_len = self.buffer.len();
>                 self.buffer.swap_remove(self.rng.gen_range(0, buffer_len))
>             }
> ```

You are right, I'll update the commit message!

twuebi (approved):

PR #162, opened 2019-11-01
Rewrite pretrain loop to support validation every N steps

### LINKED ISSUE #143: Periodical eval and summaries on tensorboard
For pretraining it's nice to do periodical evaluation and save on improvements. It's even nicer to plot this periodical evaluation in tensorboard since it nicely illustrates how the accuracy fluctuates between pretrain-batches. 

For my (non-sticker) experiments, dev-accuracy was still moving up and down by ~0.1% after ~300k train steps, although the model had reached the final level (0.02% below the highscore) after less than 200k train steps.

danieldk:
The main training loop now iterates over a given number of steps
rather than epochs. In each iteration of the loop:

- Gradient descent is performed on the number of steps specified
  by the --eval_steps option (1000 by default).
- The model is evaluated on the given evaluation set.
- If the model improves over the best previous model, the model
  is saved.

Since specifying the overal training length in terms of
steps (--steps) is often not convenient, the length can also be
specified in the number of epochs using the --epochs option.

If the training length is specified in the number of epochs, a quick
pass over the pretraining data is done to count the number of
sentences.

Currently, no Save type is used, because the interface of the savers
is currently not convenient for the pretraining scenario.

Fixes #143 

---

Sorry for the big diff + these training loops are always very imperative and messy :(.

### DIFF
--- a/sticker-utils/src/subcommands/pretrain.rs
+++ b/sticker-utils/src/subcommands/pretrain.rs
@@ -1,11 +1,12 @@
 use std::fs::File;
 use std::hash::Hash;
-use std::io::{BufReader, Seek, SeekFrom};
+use std::io::{BufReader, Read, Seek};
 use std::usize;
 
 use clap::{App, Arg, ArgMatches};
-use failure::{Error, Fallible};
-use indicatif::ProgressStyle;
+use failure::{Error, Fallible, ResultExt};
+use indicatif::{ProgressBar, ProgressStyle};
+use itertools::Itertools;
 use ordered_float::NotNan;
 use stdinout::OrExit;
 use sticker::encoder::categorical::ImmutableCategoricalEncoder;
@@ -13,13 +14,15 @@ use sticker::encoder::deprel::{RelativePOSEncoder, RelativePositionEncoder};
 use sticker::encoder::layer::LayerEncoder;
 use sticker::encoder::SentenceEncoder;
 use sticker::serialization::CborRead;
-use sticker::tensorflow::{ConllxDataSet, DataSet, TaggerGraph, TaggerTrainer};
+use sticker::tensorflow::{
+    ConllxDataSet, DataSet, LabelTensor, TaggerGraph, TaggerTrainer, TensorBuilder,
+};
 use sticker::wrapper::{Config, EncoderType, LabelerType, TomlRead};
 use sticker::{Numberer, SentVectorizer};
 
 use crate::progress::ReadProgress;
-use crate::save::{CompletedUnit, EpochAndBatchesSaver, EpochSaver, Save};
 use crate::traits::{StickerApp, StickerTrainApp};
+use crate::util::count_conllx_sentences;
 
 static EPOCHS: &str = "EPOCHS";
 static INITIAL_LR: &str = "INITIAL_LR";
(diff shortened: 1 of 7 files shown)


sticker-utils/src/subcommands/pretrain.rs
```diff
@@ -122,161 +120,180 @@ impl PretrainApp {
...
         Ok(())
     }
 
     #[allow(clippy::too_many_arguments)]
-    fn run_epoch<E>(
+    fn train_steps(
+        &self,
+        progress: &ProgressBar,
+        trainer: &TaggerTrainer,
+        steps: impl Iterator<Item = Fallible<TensorBuilder<LabelTensor>>>,
```
twuebi: rename `steps` to `batches` or something more descriptive. With `steps` I'd expect it to define how many steps this function call should perform.

danieldk: Fixed.

sticker-utils/src/subcommands/pretrain.rs
```diff
@@ -122,161 +120,180 @@ impl PretrainApp {
...
-        let mut dataset = ConllxDataSet::new(read_progress);
+        Ok((loss, acc))
+    }
 
+    #[allow(clippy::too_many_arguments)]
+    fn validation_steps(
+        &self,
+        progress: &ProgressBar,
+        trainer: &TaggerTrainer,
+        steps: impl Iterator<Item = Fallible<TensorBuilder<LabelTensor>>>,
```
twuebi: also rename to batches

sticker-utils/src/subcommands/pretrain.rs
```diff
@@ -387,33 +409,42 @@ impl StickerApp for PretrainApp {
...
+                .or_exit("Cannot parse the number of training steps", 1);
+            TrainDuration::Steps(steps)
+        } else {
+            let epochs = matches
+                .value_of(EPOCHS)
+                .unwrap()
+                .parse()
+                .or_exit("Cannot parse number of training epochs", 1);
+            TrainDuration::Epochs(epochs)
+        };
```
twuebi, line 432: Maybe either print a message here clarifying what's the precedence if both are present or use [`Arg::conflicts_with`](https://docs.rs/clap/2.33.0/clap/struct.Arg.html#method.conflicts_with) above to avoid confusion here.

danieldk: `steps` uses `overrides_with`. IIRC can't use `conflicts_with` while also specifying a default value for `epochs` (because this option will then always be present and using `steps` will always conflict). Removing the default value for `epochs` makes the interface really weird (you always have to always specify either `--epochs` or `--steps`).

twuebi: `Arg::default_value_if(STEPS,None,<defaullt val>)` may work

twuebi: This seems to work:

```rust
            .arg(
                Arg::with_name(EPOCHS)
                    .long("epochs")
                    .takes_value(true)
                    .value_name("N")
                    .help("Train for N epochs")
                    .default_value_if(STEPS, None,"2"),
            )
            .arg(
                Arg::with_name(STEPS)
                    .long("steps")
                    .value_name("N")
                    .help("Train for N steps")
                    .takes_value(true)
                    .conflicts_with(EPOCHS)
            )
```

danieldk: Thanks for sorting that out! I'll update the PR.

danieldk: No dice. For some reason `epochs` does not get a value when neither `steps` nor `epochs` is specified.

twuebi (dismissed):
Looks good to go once CI is happy.

sticker-utils/src/subcommands/pretrain.rs
danieldk: Reading the docs, `.default_value_if(STEPS, None,"2"),` seems to mean: set the value of this option (epochs) to the given value if STEPS is specified.

twuebi: My bad, I only tested whether it works without providing config and data. I guess we can keep it as before and maybe add something like "overrides STEPS" to epoch's help string.

twuebi: > Reading the docs, `.default_value_if(STEPS, None,"2"),` seems to mean: set the value of this option (epochs) to the given value if STEPS is specified.

Weirdly enough, it seems to train if `STEPS` is specified which should, in turn, trigger the default value on `EPOCHS` which, in turn, should conflict with `STEPS`.

twuebi (approved):

PR #163, opened 2019-11-04
Add the print-labels command to print a model's labels

danieldk:


twuebi:
Why take the config to retrieve the labels file instead of taking the labels file as an argument?

danieldk:
> Why take the config to retrieve the labels file instead of taking the labels file as an argument?

We need to know the encoder type to deserialize the numberer with the correct type parameter.

sticker-utils/src/subcommands/print_labels.rs
```diff
@@ -0,0 +1,81 @@
...
+        E: SentenceEncoder,
+        E::Encoding: Clone + Eq + Hash + ToString,
+        Numberer<E::Encoding>: CborRead,
+    {
+        let labels: Numberer<E::Encoding> = config.labeler.load_labels().or_exit(
+            format!("Cannot load label file '{}'", config.labeler.labels),
+            1,
+        );
+
+        for idx in 1..labels.len() {
```
twuebi, line 35: Doesn't this produce a mismatch between the number of labels in the logits vs the number of labels here?

danieldk: Could you explain this a bit further?

sticker does not currently only use 0 for padding, but it also creates the classification layer using `labels.len()`, so there first logit is used for padding. Since we never backprop padding, its probability should normally be near-0. Though, it can never be predicted, because `top_k_predictions` excludes this logit/probability.

twuebi: If I load a graph and use this label list to map into the logits, I'd be one-off due to the missing padding, right?

danieldk: That's why it prints the index as well (starting at 0). I would also be fine with printing an artificial PADDING label before looping. Would that be good enough?

twuebi: Sure, I prefer being explicit here.

twuebi (approved):

PR #164, opened 2019-11-04
Add the --maxlen option to sticker {tag, server}

danieldk:
As in training, this discards sentences that are longer than the
value given for --maxlen.

### DIFF
--- a/sticker-utils/src/sent_proc.rs
+++ b/sticker-utils/src/sent_proc.rs
@@ -13,6 +13,7 @@ where
     pipeline: &'a Pipeline,
     writer: W,
     batch_size: usize,
+    max_len: Option<usize>,
     read_ahead: usize,
     buffer: Vec<Sentence>,
 }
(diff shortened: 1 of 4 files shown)


danieldk:
Backstory: preparation of pretraining data for Dutch (apparently) failed, because some sentence contains the word *SPAM* ~6900 times, leading to 256 * 6900 overall time steps. Well, at least the sentence was accurate ;).

sticker-utils/src/sent_proc.rs
```diff
@@ -21,20 +22,34 @@ impl<'a, W> SentProcessor<'a, W>
...
             batch_size,
+            max_len,
             read_ahead,
             buffer: Vec::new(),
         }
     }
 
     pub fn process(&mut self, sent: Sentence) -> Result<(), Error> {
+        if let Some(max_len) = self.max_len {
+            // Length includes the root node.
```
DiveFish: You could add this as a note to the error message in ```server.rs```, e.g. "Cannot parse maximum sentence length (incl. root)". As a user, I would not have thought of the root being included in the sentence length.

danieldk: It should not ;). There is a bug in the next line (off-by-2) and the comment is not very clear. Thanks for catching this, I have amended the PR.

DiveFish (approved):

PR #171, opened 2019-11-11
Set the default number of warmup steps to 2000

### LINKED ISSUE #169: Proposal: set default warmup steps to 2000
We use Adam in all the models. I propose to set the standard warmup to 2000 for both `train` and `pretrain`, since it results in better models in most cases than having no warmup.

danieldk:
Motivation, Ma & Yarats, 2019:

"We conclude by suggesting that practitioners stick to linear warmup
with Adam, with a sensible default being linear warmup over
2·(1−β_2)^−1 training iterations."

We use the default Tensorflow Adam hyperparameters, where β_2 = 0.999,
2·(1−0.999)^−1 = 2000.

Fixes #169.

### DIFF
--- a/sticker-utils/src/subcommands/pretrain.rs
+++ b/sticker-utils/src/subcommands/pretrain.rs
@@ -332,7 +332,7 @@ impl StickerApp for PretrainApp {
                     .help(
                         "For the first N timesteps, the learning rate is linearly scaled up to LR.",
                     )
-                    .default_value("0"),
+                    .default_value("2000"),
             )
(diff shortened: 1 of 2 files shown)


danieldk:
cc @DiveFish : I saw that you were training a sticker model the other day. You probably want to use this changed default. The improvements for Dutch and German are ~0.6 and 0.5% LAS.

DiveFish:
Thanks for the pointer!

twuebi (approved):

PR #173, opened 2019-11-11
Improve error messages of the dependency encoders

danieldk:
The error messages now include the sentence for which processing
failed, showing the token in brackets.

### DIFF
--- a/sticker/src/encoder/deprel/relative_position.rs
+++ b/sticker/src/encoder/deprel/relative_position.rs
@@ -1,4 +1,4 @@
-use conllx::graph::{DepTriple, Node, Sentence};
+use conllx::graph::{DepTriple, Sentence};
 use failure::Error;
 use serde_derive::{Deserialize, Serialize};
 
@@ -56,21 +56,14 @@ impl SentenceEncoder for RelativePositionEncoder {
 
     fn encode(&self, sentence: &Sentence) -> Result<Vec<Self::Encoding>, Error> {
         let mut encoded = Vec::with_capacity(sentence.len());
-        for idx in 0..sentence.len() {
-            let form = match &sentence[idx] {
-                Node::Root => continue,
-                Node::Token(token) => token.form(),
-            };
-
+        for idx in 1..sentence.len() {
             let triple = sentence
                 .dep_graph()
                 .head(idx)
-                .ok_or(EncodeError::MissingHead {
-                    form: form.to_owned(),
-                })?;
-            let relation = triple.relation().ok_or(EncodeError::MissingRelation {
-                form: form.to_owned(),
-            })?;
+                .ok_or_else(|| EncodeError::missing_head(idx, sentence))?;
+            let relation = triple
+                .relation()
+                .ok_or_else(|| EncodeError::missing_relation(idx, sentence))?;
 
             encoded.push(DependencyEncoding {
                 label: relation.to_owned(),
--- a/sticker/src/encoder/deprel/error.rs
+++ b/sticker/src/encoder/deprel/error.rs
@@ -1,19 +1,97 @@
+use std::fmt;
+
+use conllx::graph::{Node, Sentence};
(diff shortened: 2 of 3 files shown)


DiveFish:
When removing the POS tag or the head of a conll token, the error message I get in training is
```Cannot read batch: cannot parse as integer field: DET```
where DET is the dependency relation of the token. Am I missing something in my test case?

File: sticker/src/encoder/deprel/relative_position.rs [hunk 1]
```diff
@@ -57,20 +57,17 @@ impl SentenceEncoder for RelativePositionEncoder {
     fn encode(&self, sentence: &Sentence) -> Result<Vec<Self::Encoding>, Error> {
         let mut encoded = Vec::with_capacity(sentence.len());
         for idx in 0..sentence.len() {
```
twuebi: Is it possible that the root has another index than `0`?

danieldk: Nope.

danieldk:
> When removing the POS tag or the head of a conll token, the error message I get in training is
> `Cannot read batch: cannot parse as integer field: DET`
> where DET is the dependency relation of the token. Am I missing something in my test case?

Did you replace the POS tag by an underscore (`_`). This error suggests that the fields have shifted and that the dependency relation is now in the head index column.

sticker/src/encoder/deprel/relative_position.rs
twuebi: Then the loop could start at index 1 and the check for root could be omitted?

```rust
for idx in 1..sentence.len() {
    ...
}
```

DiveFish:
> This error suggests that the fields have shifted and that the dependency relation is now in the head index column.

Yes, that is exactly what happens.

> Did you replace the POS tag by an underscore (`_`). 

I tried that as well - replace either the POS or the head by an underscore - but then the program simply finishes without having processed the sentence. So there is not even an error message indicating wrong input.

danieldk:
On Fri, Nov 15, 2019, at 08:21, Patricia Fischer wrote:
> >
> > This error suggests that the fields have shifted and that the
> > dependency relation is now in the head index column.
>
> Yes, that is exactly what happens.
>
> > Did you replace the POS tag by an underscore (`_`).
>
> I tried that as well - replace either the POS or the head by an
> underscore - but then the program simply finishes without having
> processed the sentence. So there is not even an error message
> indicating wrong input.

Could you send that file (or perhaps just the file with that sentence)?

Not replacing by an underscore is definitely wrong, since it shifts
the columns.

danieldk:
> Could you send that file (or perhaps just the file with that sentence)? Not replacing by an underscore is definitely wrong, since it shifts the columns.

Thanks for the example! Note that you removed coarse-grained POS. Coarse-grained tags are not used at all. In this case you can also remove the fine-grained tag, since *Diese* is not a head of any token. Try removing the tag of *lösen' and you would get:

```
Cannot collect sentence: Head of token 'Probleme' does not have a part-of-speech:

Diese [ Probleme ] lösen Studenten schnell .
```

DiveFish:
Still produces the plain error message
```Error tagging sentences: Token without a tag: lösen``` (tagging)
or 
```Cannot read batch: Token without a tag: lösen``` (training)
even after checking I am on the right branch

danieldk:
> Still produces the plain error message
> `Error tagging sentences: Token without a tag: lösen` (tagging)
> or
> `Cannot read batch: Token without a tag: lösen` (training)
> even after checking I am on the right branch

Do you use tag embeddings in your configuration? If so, then this is a different error. Since if you use tag embeddings, every token should have a tag.

sticker/src/encoder/deprel/relative_position.rs
danieldk: Fixed.

DiveFish (approved):

PR #179, opened 2019-12-04
Remove encoders, use the sticker-encoders crate

danieldk:
Still have to check that this does not introduce incompatibilities with existing models, but this should be done so that we do not have two diverging implementations of the encoders.


danieldk:
Tested with existing LSTM and transformer models. No differences in LAS, so good to go from that perspective.

twuebi (approved):

